PR #180, opened 2015-08-17
Use rust nightly

one-more-minute:
I tried building this on rust stable and it failed due to the use of feature gates.


steveklabnik:
FWIW the `run.sh` script does set up nightly, but surfacing this more clearly is probably good :+1:

PR #187, opened 2015-08-17
Fix warnings & improve idioms

steveklabnik:
This pull request does two things:

1. It fixes all warnings in regular, test, and benchmark mode.
2. It updates map.rs to use the 'tests module' pattern, which is
considered idiomatic.


jamii:
That makes so much more sense :)

steveklabnik:
:+1: 

I left the other file alone since it only has one test, there, it seems a bit much.

PR #194, opened 2015-08-17
Add some resilience to run.sh

steveklabnik:
1. This needs bash, not just regular sh. This was true before these
   changes, but also you should only use hash to test for program
   existence if you're using bash[1]
2. Test for npm and multirust appropriately.

1: http://stackoverflow.com/a/677212/24817

https://lobste.rs/s/rxwjhb/eve_version_0/comments/mz7pcn#c_mz7pcn

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,6 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Whoops, this apparently _just_ got fixed already.

PR #200, opened 2015-08-18
Fix install instruction to match actual file name

tobyjsullivan:
The installation instruction does not currently match the actual script name. This resolves it so users can copy & paste command.


ibdknox:
Looks good, thanks!

PR #201, opened 2015-08-18
Fix multirust installation command

hrishimittal:
Add missing file extension (.sh) to command.


hrishimittal:
Just noticed PR #200 already fixed this.

PR #206, opened 2015-08-18
Ensure the searcher has a larger z-index than welcome buttons

rschroll:
Currently, the "New" and "Import" buttons in the welcome message show above the searcher, if they should overlap.  This seems to be because both the buttons and the searcher are given a z-index of 10 and the buttons show up later in the DOM.  I'm guessing that there's a reason the buttons must have the z-index they do, so I've upped the z-index of the searcher-container to ensure it's comfortably over the buttons.


ibdknox:
Awesome, thanks!

Issue #169, opened 2015-08-15
Submit button in data card editor should be styled as a button


cmontella:
Especially the cursor should not be a text entry cursor

rschroll:
I think you can go ahead and close this now.

PR #207, opened 2015-08-18
Style the table "Submit" button to look like a button

rschroll:
This commit addresses issue #169, by reusing the existing .button class to style this button.

Some potential issues you may want to consider:
- In order to limit the width of the button, I had to add a parent to act as a container.  Maybe there's a better way to do this, but I'm not that familiar with the flexbox model.  (We could do it by not using flex, but it looks like you're trying to stick with flexbox everywhere.)
- The default button styling doesn't work so well against the lighter background of the form, so I copied the styles from the modal dialog for these buttons as well.  These rules are in their own section, but I'm using two selectors to pick out the two cases.  Perhaps the form and modal dialog should share a "light background" class?  Or maybe we don't want these buttons using the same styling?
- This button has a larger font size than those in the modal dialog (16px vs 13.3px).  I don't know if that's desired or not.
- I've changed the button text from "submit" to "Submit", to match the style of other buttons.

I'm happy to make changes on these or any other points you have.  Just let me know.

### DIFF
--- a/ui/css/editor.css
+++ b/ui/css/editor.css
@@ -36,6 +36,15 @@ body * { display: flex; flex-direction:column; box-sizing: border-box; outline:n
 input[type="text"], textarea { font-family: "Lato"; font-size: 16px; }
 
 /*---------------------------------------------------------
+- Buttons on lighter backgrounds
+---------------------------------------------------------*/
+
+.tabbed-box .button, .form .button { padding:8px 10px; background:#666; border-radius:3px; cursor:default; }
+.tabbed-box .button:hover, .form .button:hover { background: #777; }
+.tabbed-box .button + .button, .form .button + button { margin-left:5px; }
+.tabbed-box input[type="text"], .form input[type="text"] { border:none; background:none; border-bottom:1px solid #777; min-height:25px; padding:5px 0; }
+
+/*---------------------------------------------------------
 - Item selector
 ---------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


ibdknox:
1) Instead of adding a container, I think we can make the .form class be align-items:flex-start; and it'll do the right thing.
2) yeah, the CSS is a bit of a mess. @joshuafcole I know you're a big advocate for us creating more re-used CSS classes. Thoughts?
3) Huh, I'm not sure why those are different sizes. The modal ones should also be 16px.
4) Sounds like the right thing to do to me. :)

joshuafcole:
Making the .form class flex-start will resize all the other elements in the
.form (or rather, not resize them). I don't feel strongly on it, but I
think I prefer the straight edge on the right over a ragged one. A
container is a fine solution, but it might be more lightweight to just make
an .inline class to disable flex on a given element. `align-items: stetch`
will default fixed width children to `flex-start`. For cases where the
fixed width thing does need centered we can use `align-self: center`.

On Wed, Aug 19, 2015 at 1:55 PM, Chris Granger <notifications@github.com>
wrote:

> 1) Instead of adding a container, I think we can make the .form class be
> align-items:flex-start; and it'll do the right thing.
> 2) yeah, the CSS is a bit of a mess. @joshuafcole
> <https://github.com/joshuafcole> I know you're a big advocate for us
> creating more re-used CSS classes. Thoughts?
> 3) Huh, I'm not sure why those are different sizes. The modal ones should
> also be 16px.
> 4) Sounds like the right thing to do to me. :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/witheve/Eve/pull/207#issuecomment-132780901>.
>



-- 
Screw the environment. Please print this email immediately. And then burn
it.

rschroll:
I read over some flexbox docs again, and I realized that we can use 
align-self on the button to get the same effect as using align-items on 
the form, but without messing with any of the other children.  I've 
pushed a commit with that change.

I haven't done anything about points 2 or 3.  If you want me to fool 
with the CSS, I can.  But that might be a bit dangerous, since I don't 
know where classes are currently used or what plans are in store.  If 
you like, I can open another bug about the font size in the modal 
dialogs.

Let me know if you'd like me to squash the commits together and/or 
rebase onto master.

joshuafcole:
Yeah, 2 is a much bigger meta-issue that's probably blocked by #214. 3 is caused by .tooltip (and is itself a weird side effect of all of our modals being worked into the tooltip subsystem instead of vice versa). We can worry about fixing that later. This looks good to me, thanks for contributing.

Issue #232, opened 2015-08-20
Uncaught TypeError: a.localeCompare is not a function


rschroll:
I've managed to come up with a query that throws this error in the browser when you try to open it.  The result of the "Save to Gist" operation is here: https://gist.github.com/anonymous/80693804489f303327da

The problem is from line 89 of tableEditor.ts, where the comparison function falls back on the localeCompare method.  I've added some debugging and have found that in this case, a is `false`.

It would be easy enough to add a test here for the case of booleans, but I'm not sure how you'd like them to sort.  I also don't know if these booleans are unexpected, which might be a sign of an underlying bug.

This is seen running Eve dd3df3c7 on Ubuntu 14.04 and Chrome 44.0.

joshuafcole:
It's a bit odd, but I think the right solution is to coerce any non-numbers to string to get consistent sorting behavior, so in the default clauses when sorting rows we'd want a.toString().localeCompare(b)

Issue #231, opened 2015-08-20
We need to distinguish between different value types


jamii:
The UI so far presents the illusion that everything is just text and numbers can magically be added together. But the internal representation affects semantics through joining, sorting, grouping etc eg "19" will not join against 19.

Either we store *everything* as text and coerce only inside arithmetic, or the user needs to be able to see the difference. This affects display, data entry and importing.

cmontella:
Can we do as Excel does and choose the type on input? If it looks like a number, it's a number. We can handle the corner cases like zip codes, phone numbers, and SSNs specifically (mostly numbers that can start with a zero).

PR #236, opened 2015-08-20
Force to string before using localeCompare

rschroll:
This follows the suggestion in #232, which seems sensible to me.


joshuafcole:
We may want to revisit it once we've figured out #231, but this certainly beats the current behavior.

Thanks for your contribution.

PR #239, opened 2015-08-21
Add PhantomJS type definition

rschroll:
Without this, the run script fails with TypeScript errors from CasperJS.


rschroll:
Didn't think to check open pull requests first.  Sorry for the noise.

Issue #242, opened 2015-08-23
csv import seems to drop 0 / FALSE values 


christinac:
Try importing this file: https://www.dropbox.com/s/4z36ix9yhs9x4q8/Christina%20Cacioppo%20-%20books%20mid%202008-15.csv?dl=0

![image](https://cloud.githubusercontent.com/assets/605780/9429200/31ab6d94-497d-11e5-9e16-bd46196c433b.png)

Both the `fiction` and `fiction bool` columns show either `1`/`TRUE` or `<empty>` when they're `0`/`FALSE` in the csv.

![image](https://cloud.githubusercontent.com/assets/605780/9429188/f503802a-497c-11e5-9f15-3de27ff32397.png)

PR #243, opened 2015-08-24
Allow importing falsey values from CSV

rschroll:
The existing import code changed all falsey values to empty string.  It looks like we need to do that to undefined, but other values, like 0 or false, should be let through as-is.

This was reported in #242.


ibdknox:
Awesome, thanks! :D

PR #248, opened 2015-08-25
Clean up run.sh

hachibu:
Clean up the following in run.sh:
* Fix shellcheck warnings.
* Switch to #!/usr/bin/env bash to use the current environment's bash.
* Parse command line flags.
* Add help message.
* Use for loop to dry code up when checking dependencies.
* Send pushd and popd stdout and stderr to /dev/null to declutter script output.
* Check multirust list-toolchains for existing install before overriding.
* Store multirust version in a variable.

Minor:
* Reword a few sentences.

screen shot 2015-08-25 at 5 16 32 pm

screen shot 2015-08-25 at 3 51 04 pm

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,36 +1,72 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-waitUrl="`pwd`/ui/waiting-room.html";
+waitUrl="$(pwd)/ui/waiting-room.html";
+rustVersion="nightly-2015-08-10"
+debugFlag=false
+noBrowserFlag=false
 
-# Ensure dependencies are installed.
-echo "* Checking dependencies..."
-hash tsc 2>/dev/null
-if [ $? -ne 0 ]; then
-  echo "Please install the typescript compiler with ('sudo npm install -g typescript') before continuing."
-  exit
-fi
-hash multirust 2>/dev/null
-if [ $? -ne 0 ]; then
-  echo "Please install multirust with ('./install-multirust.sh') before continuing."
-  exit
-fi
+# Parse command line options.
+while test $# -gt 0; do
+case "$1" in
+  -h|--help)
+    echo "Compile and run the Eve editor"
+    echo ""
+    echo "Usage:"
+    echo "    run.sh [options]"
+    echo ""
+    echo "Options:"
+    echo "    -h, --help          Print this message"
+    echo "    -d, --debug         Debug build"
+    echo "    -n, --no-browser    Do not open editor in browser"
+    exit 0
+    ;;
+  -d|--debug)
+    debugFlag=true
+    shift
+    ;;
+  -n|--no-browser)
+    noBrowserFlag=true
+    shift
+    ;;
+  *)
+    break
+    ;;
+esac
+done
 
-pushd .
-  # Try using the typescript compiler (tsc) to compile UI
-  echo "* Compiling Editor..."
-  cd ui
-
-  tsc
-  if [ $? -ne 0 ]; then
-   echo "Failed to compile editor, bailing."
-   popd
-   exit
+# Ensure that dependencies are installed.
+printf "* Checking dependencies..."
+deps="tsc multirust"
+for dep in $deps; do
+  if ! which "$dep" &> /dev/null; then
+    printf "\n  x Please install %s: " "$dep"
+    if [ "$dep" = "tsc" ]; then
+      echo "sudo npm install -g typescript"
+    elif [ "$dep" = "multirust" ]; then
+      echo "./install-multirust"
+    fi
+    exit 1
   fi
-popd
+done
+echo "done."
 
-# If we aren't restarting, open the editor in the user's preferred browser
-if [[ "x$1" != "x--restart" ]]; then
-  echo "* Opening $waitUrl"
+# Try using the TypeScript compiler (tsc) to compile UI.
+pushd . &> /dev/null
+  printf "* Compiling editor..."
+  cd ui
+  tscError=$(tsc)
+  if [ $? -ne 0 ]; then
+    printf "\n  x %s\n" "$tscError"
+    popd &> /dev/null
+    exit 1
+  else
+    echo "done."
+  fi
+popd &> /dev/null
+
+# If noBrowserFlag is false open the editor in the user's preferred browser.
+if ! $noBrowserFlag; then
+  echo "* Opening editor: $waitUrl"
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -1,36 +1,74 @@
...
+    fi
+    exit 1
+  fi
+done
+echo "done."
 
-  tsc
+# Try using the TypeScript compiler (tsc) to compile UI.
+pushd . &> /dev/null
+  printf "* Compiling editor..."
```
joshuafcole: Out of curiosity, why printf instead of echo here?

run.sh
```diff
@@ -1,36 +1,74 @@
...
+  else
+    echo "done."
   fi
-popd
+popd &> /dev/null
 
-# If we aren't restarting, open the editor in the user's preferred browser
-if [[ "x$1" != "x--restart" ]]; then
-  echo "* Opening $waitUrl"
+# If openBrowserFlag is set open the editor in the user's preferred browser.
```
joshuafcole: I think we probably want to do this by default. Better to have two editor windows than zero as a first time user experience. Thoughts?

run.sh
```diff
@@ -38,18 +76,24 @@ if [[ "x$1" != "x--restart" ]]; then
...
-  # Ensure rustc is updated
-  echo "* Updating rust if necessary..."
+pushd . &> /dev/null
   cd runtime
-  multirust override nightly-2015-08-10
 
-  # Compile runtime server
-  echo "* Compiling server... (This takes a while)"
+  # Ensure rustc is updated.
+  multirust list-toolchains | grep "$rustVersion" &> /dev/null
```
joshuafcole: @jamii Do we still need to check that this is the right version of rust currently?

run.sh
hachibu: :+1:

run.sh
hachibu: @joshuafcole I agree with you. It is a better user experience for the editor to just pop up.

Maybe we can pick a better flag name that's more descriptive?

run.sh
joshuafcole: I confirmed that we need to override irrespective of whether it's installed or not. Since multiple rust versions can be installed in multi-rust, we need to tell it to use the appropriate version. I do like the clearer diagnostic output though. Also, for consistency, in the case that rust does need updated it would be nice to use printf and show completion like you are elsewhere.

run.sh
joshuafcole: Yeah, I agree. Perhaps `--headless`? Or the more verbose `--noBrowser`?

joshuafcole:
Everything else looks great by the way. Once those two things get updated I'll be happy to merge this in. Thanks for your help.

run.sh
hachibu: @joshuafcole I really like --noBrowser. What do you think about --no-browser to make it more Unix-y?

run.sh
hachibu: @joshuafcole Cool. Can you elaborate more on the case that could be more consistent?

run.sh
joshuafcole: That's fine. I should update the test.sh script to match. I left it lowerCamel to make the casperjs interop simpler, but I can paper over that on the JS side I guess.

run.sh
joshuafcole: If we detect that rust doesn't have `$rustVersion` installed, we know `multirust override "$rustVersion"` will trigger an installation. Since we echo "* Updating rust..." when this happens, and since multirust is blocking, it would be nice to printf that message and echo "done." when it completes the way you have for the other build phases.

run.sh
hachibu: :+1:

run.sh
hachibu: Ok, I understand now. Thank you. I will patch this up.

run.sh
```diff
@@ -1,36 +1,74 @@
...
+    ;;
+esac
+done
+
+# Ensure that dependencies are installed.
+printf "* Checking dependencies..."
+deps="tsc multirust"
+for dep in $deps; do
+  if ! which "$dep" &> /dev/null; then
+    printf "  x Please install %s: " "$dep"
```
joshuafcole: Perhaps we should leave the X on this line but print the fix on the next line? It looks a bit cluttered on a single line, and some people (though I can't fathom why) still use 80 character terminal widths.

screen shot 2015-08-25 at 4 56 27 pm

run.sh
```diff
@@ -1,36 +1,74 @@
...
+done
+echo "done."
 
-  tsc
+# Try using the TypeScript compiler (tsc) to compile UI.
+pushd . &> /dev/null
+  printf "* Compiling editor..."
+  cd ui
+  tscError=$(tsc)
   if [ $? -ne 0 ]; then
```
joshuafcole: `set -e` means we never get into this branch, and since you capture the output of tsc instead of letting it write to screen, we lose the failure output. I'm not sure what the right solution here is though.

hachibu: @joshuafcole I just ran into this! I think I was a little over-zealous adding set -e. I'm going to remove it.

run.sh
hachibu: :+1: I agree. I'll place it on the next line.

hachibu: @joshuafcole what do you think about this? 

screen shot 2015-08-25 at 5 13 12 pm

hachibu: I accidentally left out the sudo. But I'll put it back in.

joshuafcole: Yep, that works.

run.sh
joshuafcole: This is the last issue, then we're good to go.

hachibu: Awesome. My commit is going through CI.

hachibu:
@joshuafcole when you feel like this is good to go I can squash this into a single commit.

hachibu:
@joshuafcole I just pushed the multirust override fix. When you get a chance take another look over everything and then I can squash.

joshuafcole:
Looks great, squash when you get the chance and we'll get it merged. Thanks again for the contribution.

ibdknox:
@hachibu Hey Ray, I just wanted to say thanks for all the work here! It is greatly appreciated :)

hachibu:
@joshuafcole @ibdknox thanks guys! i'm really happy to contribute.

I just updated this branch, squashed my commits and CI has passed so it should be ready to merge.

Issue #252, opened 2015-08-28
Tooltips at the bottom of the screen are clipped


cmontella:
In particular, the settings tooltip is never fully visible

PR #253, opened 2015-08-31
Move tooltips vertically to keep entirely on screen

rschroll:
To address #252.

A few potential issues:
- Performance on Firefox can be poor, with the post-render step taking 35-40 ms.  This doesn't seem to be a problem in Chrome.  I don't understand the difference.
- If a tooltip runs off the bottom, it gets moved to align with the bottom.  Perhaps it should be moved a bit further so the drop shadow is visible?
- Only vertical position is considered.  Should we also worry about these running off the right of the screen?  This would require a bit more logic to avoid drawing the tooltip over top of the button.
- I only trigger this for things created by "showButtonTooltip".  Perhaps it should happen for everything created by "showTooltip"?  I don't understand the distinction between these two things.

### DIFF
--- a/ui/src/editor.ts
+++ b/ui/src/editor.ts
@@ -124,6 +124,14 @@ module drawn {
         }
     }
 
+    // Move the node vertically to ensure it doesn't run off the bottom on the screen.
+    function ensureOnscreen(node, elem) {
+      let maxTop = Math.max(document.documentElement.clientHeight, window.innerHeight || 0) - node.offsetHeight;
+      if(node.offsetTop > maxTop) {
+        node.style.top = maxTop;
+      }
+    }
+
 	//---------------------------------------------------------
   // Renderer
   //---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ibdknox:
I suspect the performance issue comes from forcing a layout. I think @joshuafcole wants to take a look at that first before we pull this in.

rschroll:
Yeah, I tried to avoid excessive querying of the DOM, but at some point 
we do need to find out how big the thing is.  We could try caching the 
window size, but I don't think that's a major issue.

I notice that the poor performance occurs only when the popup is first 
created, and it seems to be independent of whether the popup needs to 
be moved or not.  My inference here is that it's not our code that's 
slow, but Firefox is being slow laying out the new elements, thereby 
delaying our access to the node size and position.

joshuafcole:
Yeah, that seems plausible. The good news is, I can't reproduce the slowdown in FF Nightly on Mac (unsure which version you're using?) and 40ms is slow but acceptable in the interim for something that switches contexts like a popup. Thanks for the help!

rschroll:
For the record, I was using Firefox 40.0 on Linux.  I'll try to 
remember to test when the next version comes out to see if this gets 
better.

PR #258, opened 2015-09-02
Review master

ibdknox:


ibdknox:
@joshuafcole fix tsc stuff
@ibdknox fix old editor field ordering

Issue #96, opened 2015-08-10
Change "hidden" tables/views to "system" tables


joshuafcole:

jakedowns:
Is this still desired? Would you like the localStorage item name changed to `showSystem` as well?

jakedowns:
Also, would "show system tables" be better for the button title? or just "show system"
![](https://s3.amazonaws.com/f.cl.ly/items/0M2T1I2A3j3N1t3O2T1l/Image%202015-10-29%20at%207.19.52%20PM.png)

cmontella:
Yes, this is still something we want to do. I

jakedowns:
@cmontella Cool, thanks for the reply! Let me know if I missed anything in my PR, but the visibility toggle and item count are updating on my local copy

cmontella:
NP, we'll merge it in once we have a look at it. Thanks for the help!

PR #288, opened 2015-10-30
changed 'hidden' table tag to 'system'.

jakedowns:
...updated UI & design docs to match #96


ibdknox:
@joshuafcole are you relying on the "hidden" tag anywhere such that this would impact you?

joshuafcole:
Code looks good and the behavior seems healthy. Thanks for the contribution!

PR #297, opened 2015-12-16
Point tutorial links to right places

Whoaa512:
If this is not the right place to fix these links please point me in the direction.


joshuafcole:
Looks good, thanks for your contribution!

PR #335, opened 2016-01-30
Git hook requires /bin/bash

aidanhs:
Since it uses `[[` comparisons.


aidanhs:
It occurs to me this isn't necessary since you're not working with the stuff at the root repo level any more, closing.

PR #397, opened 2016-08-04
Add support for polygon svg elements

RubenSandwich:


ibdknox:
Looks good, thanks!

RubenSandwich:
I also used it to create minimalistic clock marks for clock.eve, but it includes a lot of magic numbers:
clockpoly

I'd be more then happy to redo it with generating the points through a function if there is interest. The reason I didn't is because it would raise the knowledge required to understand this example.

PR #422, opened 2016-08-13
Add support for runtime errors + naive divergence detector #412

### LINKED ISSUE #412: Runtime error support
Runtime errors are pretty antithetical to the Eve model, but they still can happen, and we need to be able to handle them. The primary runtime error that users run into is due to user error, in which they write code that won't fixedpoint. However, since we don't expect users to worry about our FP semantics (let alone get everything right on their first try) we need a much more gentle approach to non-converging input than hanging. 

In the long run, I think @convolvatron's suggestion of a bloom filter to detect fishy iterations that collects and compares full intermediate diffs when it hits is a clear solution to to every type of accidental N-cycle I've seen so far.

To make steps towards this, I propose a more naive cycle detector in the same format. This detector bails after a set number of iterations, and can provide debugging information on 2-cycles by running an additional iteration and capturing its output to return through the error channel.

The work required to make this happen is:
- [x] A safe path for bailing and reporting errors in the runtime
- [x] A counter based bailing case that captures the current diff and an additional iteration's diff for debugging information.
- [x] A runtime error payload in our WS messaging protocol. The message should be compatible with N-cycles so it doesn't need to be changed when the full cycle detector is added.

Future work:
- [ ] Integrate @convolvatron's bloom filter based cycle pre-filter
- [ ] Switch to full capturing mode until a repeat is found or the evaluation fixedpoints
- [ ] Modify the convergence error to support reporting N-cycles on the server side.

joshuafcole:
Fixes #412 

Remaining work items
- [ ] Client support for displaying errors
- [x] Add support for reporting additional data at error sites
- [x] Capture diffs after divergence detected to give the user debugging power
- [ ] Convert memory offsets (horribly unstable across binary changes) to a proper stacktrace (via DWARF?)


joshuafcole:
Okay, we've reached MVP for this. Errors are just splatted to the console and memory offsets aren't converted to stacktraces yet, but this is still a much better UX than "hang forever" / "crash the server"

PR #427, opened 2016-08-15
Add in line numbere

RubenSandwich:
Line numbers are not included in the included Eve editor.

Line numbers are important because error messages list line numbers as reference point to finding the location of the error. They are particularly important if someone is using their own tooling.
nolinenumbers

The editor is more pleasing to the eye without line numbers, but due to the current errors reporting line numbers I feel that they are important to include. (This could also be solved by a 'take me to this error button'.) I fooled around with the idea of only including the line numbers on the 'code' lines, but it required hacky CSS code that I was not comfortable with.


cmontella:
Ruben, I'm sorry we let this PR sit for so long without a response. The reason we haven't merged it is that we've been redoing the editor. Before we merge, we're going to see how things land, as the changes might be significant.

RubenSandwich:
No worries Corey,

I understand the alpha status of Eve and will not be hurt if pull requests are unanswered or not accepted.

Honestly the line numbers I put in are a temporary and established solution to problem of taking about the same section of code. I'm not sure they are the best solution though. I see line numbers answering the following question:

How do I refer to a specific section of code?

However I don't feel that line numbers need to necessarily be the solution to this problem.

Often times we talk about line numbers because of errors. This could be solved with a 'take me to the error' button or a error symbol next to the offending line. (Or both.)

Sometimes we talk about line numbers to point others to a location. This could be solved by showing everyones current location in the editor and a quick way to jump to their location. (You guys spoke a while ago about distributed editing.) In the case that you cannot collaborate at that specific moment perhaps you could create a link to a specific piece of code in a specific point in time. (E.g. github code links.)

However, I do think the tools should have the option of outputting line numbers just in case someone is adamant about their own development environment.

cmontella:
Agreed. So I'm going to go ahead and close this, and we'll keep you updated on the progress of the editor.

PR #430, opened 2016-08-23
Add utilities for reflecting Lua-level Eve compilation artifacts as EAVs

joshuafcole:
This is the first step in making separating and making introspectable the individual compilation phases. The immediate next step is to thread the synthetic bag through each phase of compilation and write interesting tidbits there instead of the context sideband; and to send this down to the client instead of the current pre-formed lump of JSON.

This allows us to begin work on:
- Advanced debugging tooling
- Bootstrapping compilation stages in Eve
- Integration testing on a stage by stage basis


src/db.lua
```diff
@@ -249,4 +258,386 @@ function getArgs(schema, bindings)
...
+  bag.name = args and args.name or "Unnamed"
+  bag.id = args and args.id or UUID:new()
+  if args then
+    bag:addMany(args)
+  end
+  return bag
+end
+
+-- @FIXME: no way to remove atm because we don't cache dirty removed EAVs
+function Bag:sync()
```
joshuafcole: Just sync immediately unless it becomes a problem

src/compiler.lua
```diff
@@ -971,6 +974,84 @@ function DependencyGraph.__tostring(dg)
...
+      end
+    end
+  elseif obj.type == "variable" then
+    neue.generated = obj.generated
+    neue.cardinal = sanitize(obj.cardinal, mapping)
+  elseif obj.type == "binding" then
+    neue.field = obj.field
+    neue.variable = sanitize(obj.variable, mapping)
+    neue.constant = obj.constant and obj.constant.constant
+  else -- Some kind of node
```
joshuafcole: Turn this guy into a setty loop

PR #432, opened 2016-08-26
TS-ify the client and plumb the compiler bag

joshuafcole:
This is ready to go. Of note, it can *only* work with `examples/bootstrap.eve` right now (or hypothetically, any file with its contents pasted in) since that contains the blocks necessary for providing syntax highlighting and code information to the editor.

You are expected to have `tsc` v2.0-beta installed globally with `npm i -g typescript@beta`. The Makefile updates handle the rest, so compilation is still just `make`.

screen shot 2016-08-26 at 12 45 43 am

In no particular order this PR changes:

## Client
* TS-ifies the client code
* Refactors the client into `util`, `client`, `renderer`, `editor`
* Adds AVE indexing support to the client with subscription for changes
* Rips out the old parse graph JSON support and replaces it with a new EAV based system

## Build
* Cleans up the build process for static files

## Server
* Standardizes the serving paths for static files
* Merges in the compiler bag changes that haven't hit this branch yet
* Threads the new lua compiler bag through the server
* Adds a bootstrap file in `examples/bootstrap.eve` which contains the secret sauce for the bridge

Please give it a cursory review so you're familiar with what's up when you get the chance and then merge it at your leisure @ibdknox. Ping me with any questions.


index.html
```diff
@@ -115,10 +115,22 @@
...
-    <script type="text/javascript" src="jssrc/editor.js"></script>
+    <script type="text/javascript" src="js/codemirror.js"></script>
+
+    <script type="text/javascript" src="js/system.js"></script>
+    <script>
+      SystemJS.config({
+        baseURL: "/js",
+        defaultJSExtensions: true,
+        meta: {
+          codemirror: {format: "amd"}
```
joshuafcole: I'm 95% sure I don't need this meta bit since I ended up registering it manually.

jssrc/client.ts
```diff
@@ -0,0 +1,420 @@
...
+socket.onclose = function() {
+  console.log("Disconnected from eve server!");
+}
+
+//---------------------------------------------------------
+// Bootstrapping interface
+//---------------------------------------------------------
+export var parseInfo = {blocks: [], lines: []};
+
+let updateEditorParse = debounce(handleEditorParse, 1);
```
joshuafcole: We want to call this after any change that effected entities tagged `token` or `block`, not just when new entities are added/removed from the index. We can do that by cross-referencing the `dirty` entity list with the `indexes.byTag.index[<"block" | "token">]` entity lists

PR #453, opened 2016-09-26
Fix relative paths in build/Makefile

brainopia:
The `build/Makefile` changed in a way that prevents top-level `Makefile' from running

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -70,7 +70,7 @@ contgen: contgen.c
 	cc $< -g -o $@
 
 continuation_templates.h: contgen
-	contgen 10 10 > continuation_templates.h
+	$(BUILD)/contgen 10 10 > continuation_templates.h
 
 lutf8lib.o: lutf8lib.c
 	cc -std=c99 -c -g -I$(LUAJIT_INCLUDES) $^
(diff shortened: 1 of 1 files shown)


brainopia:
Oh, my bad. You're absolutely right about `cd`. The issue is that I need any kind of path (relative or absolute) to invoke binary (duh, the `build` directory is not in my `$PATH`). Absolute `$(BUILD)` path accidentally solved my issue, but a relative `./` would fix the issue as well and it's definitely shorter :D 

I want to send later a PR with a new rule in Makefile that checks for typescript and its version (I've stumbled upon it myself, since I didn't have typescript of version 2.0). I can just as well replace my `$(BUILD)` lines with `./` in the aforementioned PR, if that's ok.

convolvatron:
its all fine. this patch is good, and i'll try to make sure that when we get it all glued back together we don't depend on having . in your path (since almost no one does that anymore)

if you have any suggestions for either making the the TS installation more clear, or at minimum 
giving better indication if that environment isn't sane, we'd love to accept them. i just spent 15 minutes this last weekend trying to figure out why 'tsc' returned an error, when it was invoking
some old x25 binary called 'node' as an interpreter.

thanks again

PR #458, opened 2016-10-07
Change dump_display use print_buffer_json to produce correctly escaped JSON strings (fixes #457)

### LINKED ISSUE #457: Newline in string literal results in Eve socket sending invalid JSON (unescaped newline)
With this program:

    ```
    match
      [#session-connect]
      string = "foo\nbar\nbaz"
      split[text: string, by: "\n" index: i token: tok]
    commit
      [#div text: "{{i}}: {{tok}}"]

Eve sends this JSON through the websocket:

    {"type":"node_graph", "head": "112", "nodes":{"112": {"id": "112", "type": "move", "arms": ["111"], "display":{"a": "r6", "b": "foo
    bar
    baz"}},"111": {"id": "111", "type": "split", "arms": ["110"], "display":{"by": "
    ", "text": "r6", "token": "r5", "index": "r4"}},"110": {"id": "110", "type": "concat", "arms": ["109"], "display":{"return": "r2"}},"109": {"id": "109", "type": "scan", "arms": ["108"], "scan_type": "eAV", "display":{"e": "r3", "v": "session-connect", "a": "tag", "sig": "eAV"}},"108": {"id": "108", "type": "sub", "arms": ["104","107"], "display":{"pass": "r1", "provides": "[ r0 ]", "ids": "[ r0 ]", "id_collapse": "true", "projection": "[ r2 ]"}},"104": {"id": "104", "type": "terminal", "arms": [], "display":{}},"107": {"id": "107", "type": "insert", "arms": ["106"], "display":{"scope": "session", "e": "r0", "v": "div", "a": "tag"}},"106": {"id": "106", "type": "insert", "arms": ["105"], "display":{"scope": "session", "e": "r0", "v": "r2", "a": "text"}},"105": {"id": "105", "type": "subtail", "arms": [], "display":{"provides": "[ r0 ]", "pass": "r1"}}}}

Which is invalid (the newline is not escaped) which causes a Javascript error: Uncaught SyntaxError: Unexpected token 
 in JSON at position xxx

frankier:


convolvatron:
i think something strange is going on here, i thought this was a failure to escape json properly. 

so maybe i don't understand the issue. why does changing the format string help? tracking the display table back to where its built in luanne.c, it looks like the value should be an 'estring', or the interned runtime string. the old format was %b, which was for the non-interned buffer based strings. you changed it to %s, which is for generic C null terminated strings. (if you ask why we have 3 different string formats i'm just going to turn red and start stuttering)

so if you don't mind working through this, i guess the first question is the issue really just unescaped quotes in json or am i really off base

thanks again for looking at this stuff, there is a lot of schmutz we need to clean out

frankier:
Oh I see now that Eve has it's own printf with %b being to do with the string type. For some reason I just assumed it was the same as bash printf where %b interprets backslashes in the inserted string. Oddly, it does at least seem to stop the console error happening for me (but causes other less visible problems I would think).

It looks like v is a buffer (aka string) rather than an estring so this new patch factors out the part of print_value_json which operates on buffers into print_buffer_json so it can be used by dump_display.

Now with the program in the bug report for the same message which previously contained invalid JSON I get:

    "{"type":"node_graph", "head": "112", "nodes":{"112": {"id": "112", "type": "move", "arms": ["111"], "display":{"a": "r6", "b": "foo\nbar\nbaz"}},"111": {"id": "111", "type": "split", "arms": ["110"], "display":{"by": "\n", "text": "r6", "token": "r5", "index": "r4"}},"110": {"id": "110", "type": "concat", "arms": ["109"], "display":{"return": "r2"}},"109": {"id": "109", "type": "scan", "arms": ["108"], "scan_type": "eAV", "display":{"e": "r3", "v": "session-connect", "a": "tag", "sig": "eAV"}},"108": {"id": "108", "type": "sub", "arms": ["104","107"], "display":{"pass": "r1", "provides": "[ r0 ]", "ids": "[ r0 ]", "id_collapse": "true", "projection": "[ r2 ]"}},"104": {"id": "104", "type": "terminal", "arms": [], "display":{}},"107": {"id": "107", "type": "insert", "arms": ["106"], "display":{"scope": "session", "e": "r0", "v": "div", "a": "tag"}},"106": {"id": "106", "type": "insert", "arms": ["105"], "display":{"scope": "session", "e": "r0", "v": "r2", "a": "text"}},"105": {"id": "105", "type": "subtail", "arms": [], "display":{"provides": "[ r0 ]", "pass": "r1"}}}}"

joshuafcole:
After some review this problem has already been addressed in a branch of the C runtime, so I'm going to close this PR. Nonetheless, thank you for bringing it to our attention!

PR #460, opened 2016-10-07
Make split[...] return the element after the last delimeter

frankier:


convolvatron:
wow. ok. great, thanks for looking at this.

it seems like my proposed fix for issuing the last split string is faulty, since i
wrote a conditional based on out being non-empty.

it seems like your solution has a similar issue.

if split("/a/b/") should have 4 matching elements? maybe i'm not reading this properly

frankier:
Whoops I forgot to allocate an empty string! That's fixed now.

>split("/a/b/") should have 4 matching elements

Yes (at least I think so). This is the behaviour of Python and Javascript split so should be in line with the expectations of people who already know them (not sure about other people). If you want the other behaviour the solution is to do the equivalent of split(strip(value, delim), delim). I guess for this reason some sort of strip should be a primitive too.

It doesn't look like there's currently a way to do write unit tests in the code base (it would probably be fairly easy to whip up a Javascript test harness with PhantomJS and run tests with TravisCI) but I've tested the following program manually and it works as expected.

```
match
  [#session-connect]
  string = "\nfoo\nbar\nbaz\n"
  split[text: string, by: "\n" index: i token: tok]
commit
  [#div class: "shouldbe5" text: "{{i}}: {{tok}}"]
```
```
match
  [#session-connect]
  string = "foo\nbar\nbaz"
  split[text: string, by: "\n" index: i token: tok]
commit
  [#div class: "shouldbe3a" text: "{{i}}: {{tok}}"]
```
```
match
  [#session-connect]
  string = "foo"
  split[text: string, by: "" index: i token: tok]
commit
  [#div class: "shouldbe3b" text: "{{i}}: {{tok}}"]
```

convolvatron:
this is great thanks, don't worry about testing it, i'll poke at it and merge

we had a harness-in-eve project going, but we found of a bunch of interesting
foundational issues have to do with writing eve in eve, external functions as 
objects, scoping, etc and by the time we were finished decided that the whole thing
needed some rework. it'll have one in the next server release. thanks again

cmontella:
Split is being done here now: https://github.com/witheve/Eve/blob/master/src/runtime/providers/string.ts#L35

PR #485, opened 2016-10-29
Correct anti-diagonal scoring in tic-tac-toe example

gamebox:
Noticed that the language - and the code - was exactly backwards for this calculation (yay narrative), fixed it.


cmontella:
Thanks for the fix, I'm glad literate programming is paying off for us! The reason for the bug is that we changed the language to be 1-indexed, an this program didn't get updated.

PR #501, opened 2016-10-31
Fix code fences in analyzer.eve

carlthuringer:
I don't know much about Eve, but opening the analyzer.eve example messes up my whole eve session in a way that's difficult to recover from.

A simple look at the source showed me that the code fence style varies wildly in there, so I performed a find and replace. Now the page loads/compiles fine, but I still don't know what it was supposed to do.

### DIFF
--- a/examples/analyzer.eve
+++ b/examples/analyzer.eve
@@ -4,36 +4,36 @@
 
 For every variable there is a group
 
-...
+```
 search
   variable = [#variable]
   number = random[seed: variable]
 commit
   [#group number variable]
-...
+```
 
-...
+```
 search
   variable = [#variable not(group)]
   group = [#group variable]
 commit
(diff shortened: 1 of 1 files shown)


cmontella:
Carl,

The `...` code fence was to make the block not execute. Since we wrote this, we've added a "disabled keyword" that you can add to a block heading to indicate that the block should not do anything. For example:


    ~~~eve disabled
    bind @browser
      [#div text: "Hello!"]
    ~~~

Will not do anything. Code fences can either be tildes or backticks, according to the CommonMark spec, so we support both.

I've updated the doc to include the new fences, so I'll close this PR. Thanks for pointing this out!

Also, the purpose of this code is for the Eve analyzer. It's part of the engine driving the inspector button.

PR #504, opened 2016-11-01
Add node id to InsertAction constructor calls.

### LINKED ISSUE #483: http example sends request twice
In the http example the request is sent twice when you press the button. In the quickstart example it says (in reference to a click event) that the record only exists for an instant but apparently it exists long enough that it gets returned twice.

vickychijwani:
Fixes #483.

Aside: Typescript should've caught these errors but didn't, so maybe there's something wrong with the project configuration?


ibdknox:
Awesome, thanks Vicky!

curioussavage:
Nice work! I didn't expect this to be fixed that fast.

PR #506, opened 2016-11-02
Add tickbox change event, propagate to records & add illustrative example

frankier:
Note that the change event deviates from the DOM change event for radio buttons since it fires on both the newly selected radio box and the old, automatically unselected radio box. This is to make it easier to sync the browser from it.

### DIFF
--- a/examples/event.eve
+++ b/examples/event.eve
@@ -30,3 +30,19 @@ update values
   commit @browser
     element.value := value
 ~~~
+
+add checked
+~~~
+  search @event
+    c = [#change element checked: true]
+  commit @browser
+    element.checked := true
+~~~
+
+remove checked
+~~~
+  search @event
+    c = [#change element checked: false]
+  commit @browser
+    element.checked := none
+~~~
--- a/examples/form-binding.eve
+++ b/examples/form-binding.eve
@@ -1,0 +1,48 @@
+Input value
+```
+commit @browser
+  [#label text: "Enter your name: " children: [#input type: "text" name: "name" value: "bill"]]
+```
+```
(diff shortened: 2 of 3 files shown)


ibdknox:
Looks good! The only thing I wonder is if checked should be a tag. Thoughts?

frankier:
I did think about that a bit. For the #change event it doesn't really matter either way since it's clearly deviating from the way the DOM events work anyway. But if we change the checked attribute to be a tag I feel like the fairly simple mapping between the Eve record based virtual DOM and the actual DOM becomes more complicated with arguably an insufficient payoff. And if we allow both we're making a distinction without a difference which is something which I think should be avoided (it increases cognitive overhead with no benefit).

On the other hand perhaps there is some benefit here to making the larger platform consistent. "Should I use a boolean or a tag to represent this flag" is probably quite a common decision which basically has little consequence either way. If there's a strong convention to prefer one or the other or follow some sort of criteria then this could free up people's time to think about more important things. I guess your preference to always use tags, or are there some criteria you have in mind? This might be useful content for a style guide.

To summarise, I went with this way since I think there's a big advantage to keeping the DOM-Eve mapping as simple as possible so people can get maximum value out of their existing skills.

vickychijwani:
There may be one more consideration here: If checked is a tag, how would you use a boolean expression to dynamically set its value? A boolean attribute looks like the right choice (though I may be missing something). And there is precedent for using boolean attribute in such a case: CSS class names are written as boolean flags.

ibdknox:
Seems like good arguments. Thanks @frankier, @vickychijwani! :)

PR #516, opened 2016-11-04
Replace angle with radians and degrees in sin and cos

### LINKED ISSUE #513: Can't use trig functions
Running this (slightly modified from http://docs.witheve.com/handbook/math/sin/):
```
search
  y = sin[degrees: 90]
  x = sin[radians: 3.14 / 2]
  
bind @browser
  [#div text: "y is {{y}}"]
  [#div text: "x is {{x}}"]
```
Result:
```
y is NaN
x is NaN
```
Same for cos.  When I try tan, I get an editor error "There's no definition for the function tan".

For what it's worth, the clock example works for me but I haven't tried editing it.

olafleur:
This pull request is to fix #513 .

### DIFF
--- a/examples/clock.eve
+++ b/examples/clock.eve
@@ -4,9 +4,9 @@ Draw an appropriately positioned line into `#clock-hand`s.
 
 ```
   search @browser
-    hand = [#clock-hand angle length]
-    x2 = 50 + (length * sin[angle])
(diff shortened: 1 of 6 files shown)


ibdknox:
Looks great, thanks!

olafleur:
My pleasure ! :)

PR #517, opened 2016-11-04
Support for query-params; New String Function: urlencode; @http bugfix

renegr:
I had the need for some more features, while implementing an OpenID Connect Client in Eve.

**About the client:**

The current state & implementation for the OIDC client can be found here:

https://gist.github.com/renegr/0dbc9b05efc4d7305fd1764770e3059d

The actual lines of code needed to implement an OIDC Client in Eve is surprisingly small 👍 

**About the pull request**

The code works, but I am not so sure if the implementation for urlencode is "proper Eve" ;)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -349,15 +349,19 @@ export function sendEvent(records:any[]) {
 
 function onHashChange(event) {
   if(_ide.loaded) changeDocument();
-  let hash = window.location.hash.split("#/")[2];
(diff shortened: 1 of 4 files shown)


ibdknox:
huh, I didn't realize the request had to be opened before setting headers, I figured it queued them. Thanks for the fix!

The rest looks good as well. :) Thanks!

renegr:
That was Quick :-) glad I could contribute.

PR #518, opened 2016-11-04
Add system-polyfills.js to src.

Cormac-Williams:
Should correct a problem in IE11 whereby SystemJS is undefined on attempting to render the main page.


ibdknox:
Awesome, thanks! :)

PR #521, opened 2016-11-05
Make CSS, favicon and fonts local

### LINKED ISSUE #520: Online icon dependencies
When disconnected from the internet, I noticed that the interface is broken and feels unusable (unless you use keyboard shortcuts I suppose), but all the icons disappear since we're using ion for icons, and we're currently importing that over a CDN. We should have our icons be a part of the package to allow the interface to look and feel the same when disconnected from the Internets.

olafleur:
This is suggested to fix #520


pranaygp:
[Author of #520 ] This seems pretty good IMO. I'll report back if there are any other online dependencies I find later while working with eve

ibdknox:
Thanks!

olafleur:
You're welcome ! :)

Issue #523, opened 2016-11-06
Pow function has arguments inverted


Cormac-Williams:
i.e

```
search @session
  value = pow[value:2, by:3]
bind @browser
  [#div text:"Pow 2 by 3 = {{value}}"]
```

Produces 9 rather than 8.

Regards,
Cormac.

Cormac-Williams:
Fixed. Can close :-)

PR #524, opened 2016-11-06
Correct Pow function having arguments inverted.

Cormac-Williams:
For issuehttps://github.com/witheve/Eve/issues/523


joshuafcole:
:+1: Good eye, everybody!

Issue #512, opened 2016-11-03
Install fails on macOS Sierra


Mark92630:
With node v6.9.1, npm v3.10.0, and the latest Eve download for Mac, the npm install command yields error messages such as:

Invalid name: "@types/commonmark"

for several different @types (one per install attempt).
[npm-debug.txt](https://github.com/witheve/Eve/files/567980/npm-debug.txt)

ibdknox:
I believe this was actually an issue with NPM, see the related HackerNews thread: https://news.ycombinator.com/item?id=12860806

Can you try again?

ibdknox:
Closing for now, if this is still an issue, please feel free to reopen it.

Mark92630:
I looked at the ycombinator thread, but it is far above my technical depth. I see 38 references to NPM, but have no idea how to decide which one(s) affect me.

Also, I don't know what you mean by "try again." Am I supposed to uninstall anything first? I re-downloaded Eve for Mac, unzipped it into a new sub-directory, and ran npm install again, with a similar error message: "Invalid name: "@types/body-parser".

Mark

Mark92630:
$ npm version
{ npm: '3.10.8',
  ares: '1.10.1-DEV',
  http_parser: '2.7.0',
  icu: '57.1',
  modules: '48',
  node: '6.9.1',
  openssl: '1.0.2j',
  uv: '1.9.1',
  v8: '5.1.281.84',
  zlib: '1.2.8' }

Mark92630:
It looks to my amateur eye as though npm isn't finding any of this @types "stuff" (see the first nine dependencies):

$ cat package.json
{
  "dependencies": {
    "@types/body-parser": "0.0.33",
    "@types/commonmark": "^0.22.29",
    "@types/express": "^4.0.33",
    "@types/glob": "^5.0.30",
    "@types/mkdirp": "^0.3.29",
    "@types/node": "^6.0.41",
    "@types/request": "0.0.31",
    "@types/tape": "^4.2.28",
    "@types/ws": "0.0.33",
    "body-parser": "^1.15.2",
    "chevrotain": "^0.14.0",
    "commonmark": "^0.26.0",
    "express": "^4.14.0",
    "glob": "^7.1.1",
    "mkdirp": "^0.5.1",
    "request": "^2.75.0",
    "typescript": "^2.0.3",
    "ws": "^1.1.1"
  },
  "devDependencies": {
    "faucet": "0.0.1",
    "tape": "^4.6.0"
[...]

PR #525, opened 2016-11-06
Move compiled files into git for easier startup

pranaygp:
After reading issue #512 , leaving '@types/xxx' dependencies as `dependencies` rather than `devDependencies` seems to be causing problems. Also, it's more semantically correct to move these to `devDependencies`. 

Though both `dependencies` and `devDependencies` are both automatically installed when running `npm install` from the project directory, people with similar issues to set up can now use `npm install --production` to only install `dependencies`.

( This PR is really just a quick response to @joshuafcole and deserves all credit for bringing happiness to the world )


pranaygp:
So, now that we have our compiled files on git, we shouldn't rebuild them on `npm install`. My recommended method here is to remove install and postInstall lifecycle hooks (you should be able to `git clone` -> `npm start`).

Instead, tsc should be run just before `build` or `build-dist`

ibdknox:
I'm not sure that this is the right approach. We still need node to run Eve at this point. We're not *really* gaining anything by packaging these - you still need all the same dependencies. The main thing we should figure out here is how the types deps aren't making it down.

@joshuafcole thoughts?

pranaygp:
@ibdknox so, the idea with this is that you still require node. You can clone this repo and run `npm start` right after and that'll work just fine since we're compiling the latest version of eve before we push to github (we can automate this).

This way, people can use eve out of the box without having to "build from source" essentially. If they want to hack on the eve runtime, they then need to grab all the dependencies ( `npm install` ) and rebuild/compile their code.

That's the goal this PR tries to achieve. (You can get up and running in no time, without having to worry about missing dependencies).

--- EXTRA STUFF ---

My alternate thoughts on this are to bundle eve into an executable binary and distribute those on the website. We can package eve to either start the server and try opening `localhost:8000` on the default browser, or bundle eve in an electron app (should be easy to do given the way eve's source is currently structured).

Thoughts?

ibdknox:
@pranaygp I think your latter suggestion is the best way to go here. If we're worried about a great double-click and run experience, it seems like doing a simple downloadable electron app would be the best strategy. One bit of work that should probably be done before we do that though is drawing a clearer line between the runtime and the editor so that it's easy to separate the two if you just want to run an application you wrote.

pranaygp:
@ibdknox Gotcha. Yup, that makes sense. Being able to have a cli tool that you can call with something like `eve test.eve`, that either runs (or preferably compiles to JS), with an electron app for the editor that uses the cli would be a good way to go with this. I do feel like this isn't high on Eve's list of priorities right now. 

If we plan on getting into that, let me know how I can help with the shift to separate the 2 parts of the experience.

In the meanwhile, is there anything we can do with the code on this PR. I do still believe having our built/compiled files in VCS makes sense. As long as someone doesn't plan on hacking on the Eve runtime or the editor itself, it makes sense to be able to clone and npm start to jump into working with Eve.

jonnor:
Opinion from a non-core contributor here. Please don't put compiled artifacts into master. It is a nightmare to work with, for instance when changing branches: Always have to reset the built files, otherwise git will not overwrite them. In merges also they are a pain. It pollutes the entire git history, and makes reading it, plus statistics/analysis trickier. Plus you very easily get "funny" situations where the built files and source are mismatching. Someone forgets to commit built files, or commits built files from a slightly older build.
And since user needs node/NPM anyways, it only eliminates having to run one command (npm install).

Very positive to having a downloadable executable though!

ibdknox:
Yeah I don't think we should put the compiled output in the repo. We've done that before and as @jonnor said, it ended up being a bit of a nightmare.

pranaygp:
@ibdknox Fair enough. A lot of those concerns make sense to me. We should focus on making an executable in the future after a lot of Eve's fundamentals have been figured out.

tbodt:
@ibdknox Will eve ever not require node to run?

pranaygp:
@tbodt If (or when) we make an electron version of Eve, you can run it locally without node. The best version to currently use Eve without node is at  http://play.witheve.com

ibdknox:
@tbodt In the longer term it won't need node to run. We intend to bootstrap more and more of it over time such that only requires a very small executor core.

tbodt:
@ibdknox Will it still require some kind of JavaScript environment?

ibdknox:
Once we're at that point, the executor can be written in anything really, so nope. We've built a prototype of this in C and with some work there's no reason something like that couldn't work most anywhere (embedded, phones, etc).

yoursdearboy:
If I understand correctly - since package is published to `npm` with compiled sources, then typings could be moved to `devDependencies`. Just noticed this while looking at source code.

PR #526, opened 2016-11-06
fix Math.pow function call having arguments in wrong order

pranaygp:
Suggested fix to #523


olafleur:
Thanks for the fix. However, this is a duplicate of #524 :)

pranaygp:
True. I did not see that PR. I can close this in favor of merging that.

PR #532, opened 2016-11-09
Add coordinates to click and dblclick events for SVG and root elements

frankier:


cmontella:
It looks like you're saving two sets of coordinates? For example, the given code produces the following output when an SVG is clicked:

~~~
search @event
  [#click x y]
  
search 
  last-click = [#last-click]

commit
  last-click.x := x
  last-click.y := y
~~~

~~~
search
  not([#last-click])
  
commit
  [#last-click x: 0 y: 0]
~~~

~~~
search
  [#last-click x y]
  
bind @view
  [#value | value: "({{x}}, {{y}})"]
~~~

Output: 

~~~
(1132, 135)
(1132, 38.33333206176758)
(24.66666603088379, 135)
(24.66666603088379, 38.33333206176758)
~~~

frankier:
Yeah. I can see how that's not so convenient from that example. The way I made it at the moment is SVG elements set it to the position in their local coordinate space and the root element sends clientX and clientY. All I actually want/need myself is the SVG coordinates but I thought document coordinates might be useful to someone too. Perhaps they should be relative to the application container however?

One way I could make your specific example work is by their names to svgx, svgy and clientx, clienty or similar. There's still the possibility of a similar issue arising in the presence of nested SVGs. The best solution is to specify the element you're interested in.

Thoughts?

cmontella:
Okay, I understand your intention now. I think maybe this is okay then. Will merge.

PR #543, opened 2016-11-12
Fix HTTP server example

jonnor:
Was looking into how to add database/providers on serverside, and the closest example didn't quite work :)


convolvatron:
this is all fine.  but for this to be useful we need to support a mode where some eve code is running up front in order to receive events in the server database, which we should do at some point really.

jonnor:
Agreed completely on needing a way to load code into server. I did not see an issue for it, so filed I it here: https://github.com/witheve/Eve/issues/559

PR #553, opened 2016-11-14
Better messages for Eve server

cmontella:
Displays a better message when the server stops, and a better error when the server cannot start due to the port being in use.

### DIFF
--- a/src/runtime/server.ts
+++ b/src/runtime/server.ts
@@ -32,7 +32,7 @@ const contentTypes = {
 }
 
 const BROWSER = !argv["server"];
-const PORT = process.env.PORT || 8080;
+const PORT = process.env.PORT || argv["port"] || 8080;
 const serverDatabase = new ServerDatabase();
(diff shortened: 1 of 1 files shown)


src/runtime/server.ts
```diff
@@ -162,5 +162,13 @@ let wss = new WebSocketServer({server: server});
...
 
 server.listen(PORT, function(){
-  console.log("Server listening on: http://localhost:%s", PORT);
+  console.log(`Eve is available at http://localhost:${PORT}. Point your browser there to access the Eve editor.`);
+});
+
+// If the port is already in use, display an error message
+process.on('uncaughtException', function(err) {
+    if(err.errno === 'EADDRINUSE') {
+      console.log(`ERROR: Eve couldn't start because port ${PORT} is already in use. Close the process using this port and try again.`);
```
ibdknox: Instead of telling them to close the thing on that port, we could tell them to use the PORT env variable to use a different port.

PR #556, opened 2016-11-14
all messages to the server should be an object

ibdknox:
The "ping" message was causing the server to crash because it isn't an object.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -214,7 +214,7 @@ export class EveClient {
     // ping the server so that the connection isn't overzealously
     // closed
     setInterval(() => {
-      this.socketSend("\"PING\"");
+      this.socketSend(JSON.stringify({type: "ping"}));
     }, 30000);
   }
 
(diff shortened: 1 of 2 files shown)


joshuafcole:
Might be worth changing the throw to a console.error while we're at it.

ibdknox:
@convolvatron the ping is there because inactive sockets are closed after a period of inactivity on a lot of hosts (e.g. heroku). So we have to send a message to the server every once in awhile to make sure that the socket stays open even if the user isn't doing anything on the page at the time.

joshuafcole:
I think this is a holdover from issues we had previously where the browser would occasionally kill websockets that had been quiet too long. I don't know that we ever figured out specifically why, but this was an empirical solution.

convolvatron:
as usual i'm confused about the underlying truth, but i think that the ws protocol explicitly
send periodic web socket level pings, and if those aren't flowing then the remote peer or 
the OS might decided to just give up. so it shouldn't(?) be necessary to have keepalives
above ws..but i don't know that, and its entirely possible that Heroku went to extreme effort 
to try to filter out the ws level pings

ibdknox:
Yeah they do something extra: https://devcenter.heroku.com/articles/http-routing#timeouts

PR #557, opened 2016-11-15
Update attributions

cmontella:
Please look over these and let me know if anything you pulled in is missing.

### DIFF
--- a/src/runtime/server.ts
+++ b/src/runtime/server.ts
@@ -32,7 +32,7 @@ const contentTypes = {
 }
 
 const BROWSER = !argv["server"];
-const PORT = process.env.PORT || 8080;
+const PORT = process.env.PORT || argv["port"] || 8080;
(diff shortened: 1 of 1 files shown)


cmontella:
When this is in, I'd like to suggest a sub release to 0.2.1, and make 0.2.0 unavailable, since these were not included and should have been.

joshuafcole:
There's typing from TSD (old types repo) in the typings directory for CodeMirror. It's unfortunately not stock anymore because it had several issues that had to be fixed which I haven't pushed upstream yet, and because of the way it's laid out I can't ambiently merge the types for other addons we use in, so I had to hardcode them in. Regardless, we still need to attribute it. 

Looks like we're also missing an attribution for System.js, which we use for loading the JS.

We also use two codemirror addons, simplescrollbars and annotatescrollbar

PR #558, opened 2016-11-15
Update attributions

cmontella:


joshuafcole:
There's typing from TSD (old types repo) in the typings directory for CodeMirror. It's unfortunately not stock anymore because it had several issues that had to be fixed which I haven't pushed upstream yet, and because of the way it's laid out I can't ambiently merge the types for other addons we use in, so I had to hardcode them in. Regardless, we still need to attribute it.

Looks like we're also missing an attribution for System.js, which we use for loading the JS.

We also use two codemirror addons, simplescrollbars and annotatescrollbar

joshuafcole:
Oh, ionicons too!

cmontella:
Okay, I added those missing attributions. Any others @ibdknox?

ibdknox:
not that I'm aware of, why does this also include the port changes though?

joshuafcole:
The typings for CodeMirror are from TSD (the DefinitelyTyped repo here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/03b3450d08a0b0a1b9e820e581a52c8e6c21f32e/codemirror/codemirror.d.ts) not from npm, since they had to be modified.

joshuafcole:
After that you're good to go.

joshuafcole:
@ibdknox the ports changes are part of the ghost commit I mentioned yesterday that resulted from separating the two feature branches with a force push after the first branch had already somehow gotten merged.

cmontella:
Okay, final change is in.

ATTRIBUTIONS.md
```diff
@@ -1,42 +1,223 @@
...
+
+### @types/ws 
+- TypeScript definitions for ws
+- By Paul Loyd (https://github.com/loyd)
+- [MIT][MIT] License
+- https://www.npmjs.com/package/@types/ws
+
+-------------------------------------------------------------------------------
+
+### @types/codemirror
```
joshuafcole: Let's just call this `codemirror.d.ts` since it's not part of the `@types` repo. Sorry for all the nitpicking! 🙏

PR #562, opened 2016-11-16
Fix CRM app

cmontella:
The fake times in test data were causing problems with ordering in conjunction with messages with real times.

### DIFF
--- a/examples/CRM.eve
+++ b/examples/CRM.eve
@@ -25,8 +25,8 @@ search
   app = [#app]
   [#time day month year hours minutes seconds ampm]
   not(app = [#init])
-  //user = [#user name: "Corey Montella"]
-  user = [#user name: "Eric Hoffman"]
+  user = [#user name: "Corey Montella"]
+  //user = [#user name: "Eric Hoffman"]
 
 commit
   app <- [#init user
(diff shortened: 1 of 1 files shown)


examples/CRM.eve
```diff
@@ -170,15 +170,15 @@ search @browser @session
   msgs = thread.messages
   name = if msgs.sender.name = app.user.name then "Me"
          else msgs.sender.name
-  msgs.time = [hours minutes ampm]
+  msgs.time = [timestamp time-string]
```
convolvatron: this is an object without a tag?

examples/CRM.eve
```diff
@@ -399,25 +399,25 @@ commit
 
   // Make some threads
   [#thread #new-messages users: (corey, chris) messages:
-    [#message sender: corey, time: [day month year hours minutes seconds: 0 ampm], text: "Hey"]
-    [#message sender: chris, time: [day month year hours minutes seconds: 1 ampm], text: "Hey, how are you."]
-    [#message sender: corey, time: [day month year hours minutes seconds: 2 ampm], text: "I'm fine, how are you?"]
-    [#message #unread sender: chris, time: [day month year hours minutes seconds: 3 ampm], text: "Fine as well."]
-    [#message #unread sender: chris, time: [day month year hours minutes: minutes + 1, seconds: 4 ampm], text: "Glad we got that out of the way!"]
-    [#message #unread sender: chris, time: [day month year hours minutes: minutes + 1 seconds: 5 ampm], text: "What did you work on yesterday?"]]
+    [#message sender: corey, time: [timestamp: t1, time-string], text: "Hey"]
```
convolvatron: this seems fine otherwise. generally we've had the discussion about whether we want to have a time type, and even if we don't, it seems like the time string shouldn't really be carried around separately? a timestamp should stand on its own and we should have something (or more than one thing) that formats it 

maybe i am just lost?

cmontella: Right, we have no conversions right now though. I'm carrying around the timestamp to sort it, and the time-string to display it. Ideally, the timestamp would be carried around, and the time-string could be inferred.

examples/CRM.eve
cmontella: Yeah

Issue #566, opened 2016-11-17
Need a set of time conversion functions


cmontella:
Eve needs functions to convert different time representations e.g. a timestamp into a [day month year ...], or into a string e.g. October 10th 2016.

olafleur:
Cool. Do you have a suggested API for this?

For example , something like this? (throwing ideas here - not sure what's the naming convention)

```
date = toDate[timestamp: 1479353359]

datestring = toDateString[date: [day: 16 month: 11: year: 2016]]
```

cmontella:
Well, we have plans for a general convert function. This could fit in there. For instance

```
converted-value = convert[value, from: "time-stamp", to: "time-string"]
```

Which could be used in other places as well:

```
converted-value = convert[value, from: "meters", to: "feet"]
```

cmontella:
I think the convert function is general is a good idea, but we need to think more about converting time formats, because it's apparently very hairy. The last thing we want to do is support our own time conversion library, so we'll need something already that conforms to time conversion standards.

PR #577, opened 2016-11-19
WIP : more general convert function

olafleur:
This is a work in progress following a discussion in #566 .

I've added the meters <-> feets example (and tests to prove the behavior).
For the moment, I don't have other targets (would we like km? cm?) so I'm cheating and ignoring "from".

Will work on a cleaner way (and with less if/else).

Also, next on the line : timestamp / date / etc.

Waiting for your input on the wanted possibilities of conversion. ;)

### DIFF
--- a/src/runtime/builder.ts
+++ b/src/runtime/builder.ts
@@ -7,6 +7,7 @@ import * as providers from "./providers/index";
 import "./providers/math";
 import "./providers/logical";
 import "./providers/string";
+import "./providers/convert";
 import * as errors from "./errors";
 import {Sort} from "./providers/sort";
 import {Aggregate} from "./providers/aggregate";
--- a/src/runtime/providers/convert.ts
+++ b/src/runtime/providers/convert.ts
@@ -5,6 +5,7 @@ class Convert extends Constraint {
   static AttributeMapping = {
     "value": 0,
     "to": 1,
+    "from": 2
   }
   static ReturnMapping = {
     "converted": 0,
@@ -12,10 +13,11 @@ class Convert extends Constraint {
 
   resolveProposal(proposal, prefix) {
     let {args, returns} = this.resolve(prefix);
-    let from = 0;
     let value = args[0];
     let to = args[1];
+    let from = args[2];
     let converted;
+
     if(to === "number") {
       converted = +value;
       if(isNaN(converted)) throw new Error("Unable to deal with NaN in the proposal stage.");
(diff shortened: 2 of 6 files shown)


Cormac-Williams:
Hi Olafleur,

Sorry for not getting a chance to have a look earlier. Been one of those weeks :-) My open source time is limited.

Comments

- "feet" is the plural of "foot" in the archaic measurement system. There is no such word as "feets" :-)

- You are declaring a "from" attribute. This is then never used.

 I assume when you talk of removing all the if statements and the declaration of the from attribute that you are intending the convert function to wrap a general data structure that holds from,to and conversion factor information. This is certainly the way I would go.

You may want t take a look at the Frink programming language, the guy behind that has put a huge amount of work into correctly converting from one thing to another. And his data file is fun to read for geeks :-) , I'm sure I saw it mentioned in some way Eve related.
 
Regards,
Cormac.

olafleur:
Ha! Ha! You are right. Will fix feets -> feet. :)

Yes, for the moment I'm not using the "from" attribute because with the possibilities that I have, I can alway infer what it is. However, the day that we will have conversions from feet to kilometers, for example, a "from" is essential.

Yes, my intent is to have a more general structure. Not sure what is the better way though.
Maybe an abstract class with childs implementing each concrete possibility?

cmontella:
@olafleur great work, this is the kind of interface we're looking for with a convert function, and I'm glad to see it works in practice.

We had a discussion about this in the office today about the best way to do convert, since it seems like such a natural fit to be implemented in Eve itself. Let's say a user wants to implement a custom conversion. If we implement `convert[]` in Eve, the user would just add a record to the database specifying the units and the conversion ratio. Awesome. But if `convert[]` is implemented in TypeScript, the user would have to add a line to a probably very large if-statement in the source code, and then rebuild the entire Eve runtime. Not so great (as an aside, this is a perfect example of how Eve makes something incredibly simple compared to other languages).

Of course this is problematic, since we don't yet have a way for users to define functions in Eve. We spent some time talking about that as well, and we've agreed that UDFs have some tricky implications and edge cases we need to think about more before we can feel safe about committing to any one implementation.

So I think for the time being at least, you should hold off working on this more, at least until we reach a proposal for UDFs. In the meantime, I'll leave this PR open, at least to track discussion on this topic.

Any other thoughts?

olafleur:
You are right. It would be a lot more awesome if we find a way to implement it in Eve itself ! :)

Also, I'm not sure what UDF means. (Universal Dynamic Functions ? ;) )

cmontella:
Sorry, user defined function.

cmontella:
I'm going to close this PR. and we'll track the progress of the convert function here: https://github.com/witheve/Eve/issues/610

Issue #539, opened 2016-11-12
Numeric functions shouldn't return NaN


cmontella:
A function that would return NaN should instead return no results.

Cormac-Williams:
But the question becomes then, how does Eve allow the detection of out of range function use?

Or equally, how does Eve handle divide by zero? 
At the moment, whatever the result that comes out of Javascript is republished. This I agree is probably an issue.

Cormac-Williams:
I got curious so had a play. This push
https://github.com/Cormac-Williams/Eve/commit/8716ca1beb1b4b8a3baf9b66b049d139179e1628 appears to do what you wish, I essentially emulated what is already present in Logical.ts. Let me know if it works for you and I'll make it a formal pull request. 

This code now would not run

search
  div = 1 / 0
  value = "Works"
bind @view
  [#value value:"{{value}}"]

But this code runs correctly

search
  a = 1
  b = 0
  value = if not( a / b ) then
          "Ooops, divide by zero?" else
          a / b
bind @view
  [#value value:"{{value}}"]

So code written carelessly in terms of a divisor or testing for NaN is more than likely going to filter, but still leaving the possibility open of a more nuanced examination if necessary.

joshuafcole:
This has been in since 0.2.2, thanks again!

PR #589, opened 2016-11-21
Resolve NaN issue in Math with a few tests.

Cormac-Williams:
Should resolve https://github.com/witheve/Eve/issues/539


test/math.ts
```diff
@@ -53,4 +53,103 @@ test("Should be able to use the cos function with degrees and radians", (assert)
...
+    ~~~
+  `);
+  assert.end();
+});
+
+test("Divide by zero should return nothing.", (assert) => {
+  let expected = {
+    insert: [],
+    remove: [],
+  };
```
olafleur: I would suggest adding `errors: true` here to make sure that we effectively have an error.

test/math.ts
```diff
@@ -53,4 +53,103 @@ test("Should be able to use the cos function with degrees and radians", (assert)
...
+    ~~~
+  `);
+  assert.end();
+});
+
+test("ACosh < 1 should return nothing.", (assert) => {
+  let expected = {
+    insert: [],
+    remove: [],
+  };
```
olafleur: same here

test/math.ts
```diff
@@ -53,4 +53,103 @@ test("Should be able to use the cos function with degrees and radians", (assert)
...
+    ~~~
+  `);
+  assert.end();
+});
+
+test("ATanH < -1 and > 1 should return nothing.", (assert) => {
+  let expected = {
+    insert: [],
+    remove: [],
+  };
```
olafleur: here also

src/runtime/providers/math.ts
```diff
@@ -473,4 +483,4 @@ providers.provide("ln10", LN10);
 providers.provide("log2e",LOG2E );
 providers.provide("log10e",LOG10E );
 providers.provide("sqrt1/2", SQRT1_2);
```
olafleur: What is `sqrt1/2` ?

olafleur: Right. As in `i` ! :)

olafleur: Yeah. I am confused. You are right.

test/math.ts
Cormac-Williams: Lol, I tried that and pushed too early. errors: true seems to cause the test to expect an error to be raised in some way from the code. I reverted, and may need to receive guidance from the team.

Filtering NaN is a constraint, that causes the block to filter any results rather than a specific error condition that needs to be raised.

ibdknox: Yeah, the behavior you have is correct here. What we want to verify is that if anything would otherwise return NaN that it instead returns nothing. It's not an error, it's the absence of a result.

src/runtime/providers/math.ts
```diff
@@ -24,10 +26,13 @@ abstract class TotalFunctionConstraint extends Constraint {
 
   // Total functions always have a cardinality of 1
   getProposal(tripleIndex, proposed, prefix) {
-    let proposal = this.proposalObject;
-    proposal.providing = proposed;
-    proposal.cardinality = 1;
-    return proposal;
+    if(this.returns.length) {
```
ibdknox: Out of curiosity, why are you checking if there's a return?

src/runtime/providers/math.ts
```diff
@@ -63,6 +68,11 @@ function degreesToRadians(degrees:number){
 }
 
 class Add extends TotalFunctionConstraint {
+  resolveProposal(proposal, prefix) {
```
ibdknox: Why override resolveProposal? Seems like add should have the same NaN guard.

Cormac-Williams: At the moment the add function in the Math file is also providing the ability to concatenate strings to the language. If the decision is that Eve should always use string interpolation then this can be removed.

My opinion is that concatenating strings with a + symbol is such second nature to programmers that it would provide a real WTF moment to filter a block anytime "" + "" is used.

src/runtime/providers/math.ts
Cormac-Williams: To bring the code in line with the way that providers\logical.ts handles the same situation.
There's a slight discrepancy with the documentation in that it isn't returning a proposal with cardinality zero, but I've been unable to show that simply returning produces any strange behavious, so assumed returning an undefined is handled correctly by the runtime.

test/math.ts
olafleur: :thumpsup:

src/runtime/providers/math.ts
ibdknox: A fair number of languages have explicit concatenation operators instead of overloading + for strings. The main argument against doing so is that it doesn't make a lot of sense without implicit conversions which is something we want to stay far away from. E.g. "foo" + 1 implicit converts 1 to a string in JS. I think we should do more than just fail silently in the case where you try to add strings, but we need to think a bit about what that warning system would be.

src/runtime/providers/math.ts
ibdknox: Ah the logical providers do this for a reason that doesn't really apply here. There are two cases for the logic operators, the standard filtering case `x > y` and the value returning case using `is(x > y)`. In the former case, we never want to propose anything and instead just go through the testing phase. In the latter case, `is` adds a return variable that we'll want to fill in with the boolean value of the expression. With math ops, they should always have a bound return, so the check doesn't buy us anything.

Cormac-Williams: Ok, I'll revert the getProposal function to original.

src/runtime/providers/math.ts
Cormac-Williams: At the moment you don't fail silently, you add them using Javascripts implicit rules :-) . I didn't want to introduce a breaking change to other peoples Eve code, but I agree that some thought needs to be given to implicit conversion. What about a combine aggregate?

I'll leave this as is for the pull unless told otherwise.

src/runtime/providers/math.ts
```diff
@@ -13,7 +13,9 @@ abstract class TotalFunctionConstraint extends Constraint {
   // proposed variable.
   resolveProposal(proposal, prefix) {
     let {args} = this.resolve(prefix);
-    return [this.getReturnValue(args)];
+    let result = this.getReturnValue(args);
+    if (isNaN(result) || !(isFinite(result))) {return [];}
```
ibdknox: Inifinity is a well-defined value, so I think we can remove that check and just have the NaN one.

src/runtime/providers/math.ts
Cormac-Williams: Done.

src/runtime/providers/math.ts
Cormac-Williams: Removed the check. This was also was what was handling the filtering for Divide by zero, so handled that in Div function.

ibdknox:
Looks great, we just need to fix the merge conflict in math.ts and in she goes :)

Cormac-Williams:
Done :-)

ibdknox:
Huh, looks like something went sideways on the merge - there are actually more conflicts now and  @cmontella's look like they maybe got overwritten?

Cormac-Williams:
Ouch, well that's not as intended.
I'm going to revert the conflict merge.
Pull from master.
Apply changes again.
Hope and pray :-)

Cormac-Williams:
Ok, that went even more sideways. Sometimes I hate git :-(

I'm going to kill this pull request and send afresh with just the changes wanted. There'll soon be more commits than lines changed ;-)

cmontella:
Thanks, sorry for the issues.

Cormac-Williams:
Not your fault, my muckups.

Recreated as https://github.com/witheve/Eve/pull/629.

Issue #478, opened 2016-10-29
Decoding `null` in JSON throws an error


xymostech:
When trying to decode JSON that has a `null` value in it from a `[#request]` in `@http`, I'm seeing a JavaScript exception being thrown [here in eavs.ts](https://github.com/witheve/Eve/blob/master/src/runtime/util/eavs.ts#L29).

A solution would be to just check for `null` right before that line. I don't know if there's a way to represent `null` in Eve though; if not then maybe it makes sense to leave off that value from the constructed record? Or maybe there should be an explicit `null` value to represent this case, so that `toJs(fromJs(json)) == json`?

boneskull:
fwiw `typeof null === 'object'"

joshuafcole:
While we dont plan to support null as a first class value in eve, we intend
to add a  "#json-null" record or similar to the @http bag to maintain
idempotence  when reserializing.

PR #593, opened 2016-11-22
Fix error when JSON value is `null`

cgarvis:


cmontella:
Fixes: https://github.com/witheve/Eve/issues/478

xymostech:
As mentioned in #478, this would mean that the `null` value doesn't get recorded in the record at all, so de-serializing then serializing again might lose keys. I would love this merged though because it's painful not being able to use APIs that have `null`s in the responses.

ibdknox:
Yeah, there's a follow up PR that needs to happen after this one that adds the sentinel value for json nulls.

ibdknox:
Thanks!

PR #598, opened 2016-11-23
Npm package

joshuafcole:
There's still work to be done here, but we should be at the point now where the missing features are features we haven't had yet. Namely, the editor can't really work in external projects just yet, and multiple workspaces aren't supported in the UI.

Outstanding issues:

### Blocking
- [x] Need a permissions pass to serve files from global installation.
- [x] support editor on currently loaded external file
- [x] load workspace for external projects
- [x] don't crash on invalid doc urls
- [x] actually initialize in all the weird combinations of local/global/internal/external/url

### Non-Blocking

- [x] Existing CLI filepath logic doesn't work unless filepath comes before flags
- [x] Need a `eveSource.find` to find a URI in any workspace. https://github.com/witheve/Eve/blob/npm-package/src/runtime/server.ts#L184
  * Without this serving multiple documents externally doesn't work.
- [ ] Multiple workspaces support missing from the UI
  * The last registered workspace will be the only one shown.
  - [ ] https://github.com/witheve/Eve/blob/npm-package/src/ide.ts#L309
  - [ ] https://github.com/witheve/Eve/blob/npm-package/src/ide.ts#L2095
  - [ ] https://github.com/witheve/Eve/blob/npm-package/src/ide.ts#L2111
  - [ ] https://github.com/witheve/Eve/blob/npm-package/src/ide.ts#L2165

- [ ] Should move play.witheve to run off the tgz generated by npm rather than a separate packaging script.

### Maintenance

- [ ] URL control is spaghetti. https://github.com/witheve/Eve/blob/npm-package/src/client.ts#L276
- [x] Unnecessary (?) file writes when serving. https://github.com/witheve/Eve/blob/npm-package/src/runtime/server.ts#L108


src/runtime/server.ts
```diff
@@ -87,77 +69,92 @@ class HTTPRuntimeClient extends RuntimeClient {
...
-  let url = request['_parsedUrl'].pathname;
-  fs.stat("." + url, (err, result) => {
-    if(err) {
+  app.get("/build/workspaces.js", (request, response) => {
+    let packaged = eveSource.pack();
+    response.setHeader("Content-Type", `application/javascript; charset=utf-8`);
+    response.end(packaged);
+  });
+
+  app.get("/assets/*", handleStatic);
```
ibdknox: did we check to make sure this works?

src/runtime/server.ts
```diff
@@ -87,77 +69,92 @@ class HTTPRuntimeClient extends RuntimeClient {
...
 // Websocket
 //---------------------------------------------------------------------
 
+
+
 class SocketRuntimeClient extends RuntimeClient {
   socket: WebSocket;
 
-  constructor(socket:WebSocket, withIDE = true) {
+  constructor(socket:WebSocket, inEditor:boolean) {
```
ibdknox: since withIDE is used on the client, let's be consistent

src/client.ts
```diff
@@ -272,10 +274,20 @@ export class EveClient {
       browser.init(data.code);
     }
     if(this.showIDE) {
+      // Ensure the URL bar is in sync with the server.
+      // @FIXME: This back and forth of control over where we are
+      // is an Escherian nightmare.
```
ibdknox: let's turn this into issue to track

joshuafcole:
@ibdknox please do a follow-up review when you have a second, but I believe this is ready to go.

src/client.ts
```diff
@@ -464,10 +462,10 @@ function changeDocument() {
   let ide = client.ide;
   // @FIXME: This is not right in the non-internal case.
   let docId = "/examples/quickstart.eve";
-  let path = "/" + location.hash.split('?')[0].split("#/")[1];
-  if(path) {
+  let path = location.hash && location.hash.split('?')[0].split("#/")[1];
+  if(path && path.length > 2) {
```
ibdknox: why > 2?

joshuafcole: In order to be remotely valid it needs to contain at least a character and a slash. Honestly it should be 3, since all files must at least be <workspace>/<file>

ibdknox: probably worth putting a comment in or maybe a regex or something?

PR #602, opened 2016-11-26
Add missing HTML5 and SVG tags

frankier:


deltaidea:
Wouldn't it be better to move the tag list into a separate file? Possibly JSON. This looks more like a bunch of data than code. Harder to read actual logic this way.

joshuafcole:
This is tough to read, but it is done for performance reasons. A more
extensible solution unfortunately reduces performance by an order of
magnitude or more. It may be worth generating this file from a template
though to make it more manageable.

PR #603, opened 2016-11-27
Add #time.timestamp to attributeOrdering, fixing problems where [#time timestamp] searches don't get updated

frankier:


ibdknox:
Thanks!

PR #611, opened 2016-11-29
Remove unnecessary condition in fullScan

brainopia:
There are two reasons to remove this condition.

1. It's not used. All conditions when `attribute` is present are
already coverered via index lookup in other branches of switch in `getProposal`
(`*a**`,`ea**`,`*av**`,`eav*`,`***n`).

2. It's broken. Default branch of switch (which corresponds to full scan) in
`getProposal` always populates `providing` in order of EAV, but if this
condition would ever run it would return values in order of AVE thus
breaking the assumptions of `joinRound` function.

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/611)

### DIFF
--- a/src/runtime/join.ts
+++ b/src/runtime/join.ts
@@ -184,11 +184,7 @@ export class Scan {
     let solving = [];
     let solveNode = this.node !== undefined;
     let depth = solveNode ? 4 : 3;
-    if(a !== undefined) {
-      this._fullScanLookup(index.aveIndex, solving, results, [a,v,e,node], 0, 0, depth);
-    } else  {
-      this._fullScanLookup(index.eavIndex, solving, results, resolved, 0, 0, depth);
-    }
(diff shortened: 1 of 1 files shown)


brainopia:
By the way, this PR is a by-product of a work on another PR. Currently in Eve, `lookup` in `search` section is broken unless `record`,`attribute` and `value` are all present in code.
Example, 

```
  prepare
  ~~~
    commit
      [#foo]
  ~~~

  test
  ~~~
    search
      lookup[record attribute]
  ~~~
```

If you run it in shell you'll get

```
prefix[currentProvide.id] = value[providingIx];
                                     ^
TypeError: Cannot read property 'id' of undefined
    at joinRound (/home/brainopia/code/sources/eve/build/src/runtime/join.js:709:38)
```

The missing parts of `lookup` are propagated as `undefined` to `Scan` and it breaks all assumptions of current logic. I have a plan to tackle this and half-way through implementing it.

PR #613, opened 2016-11-30
Fix not without dependencies

ibdknox:
Currently nots that have no dependencies never get an opportunity to run since they are never solving for a variable. This PR adds a prejoin flag to accept and changes NotScan to force an accept check in the case when prejoin is true and the scan has no args. It also adds an extra step to the preJoinAccept function to check nots if they would otherwise not be.

### DIFF
--- a/src/runtime/join.ts
+++ b/src/runtime/join.ts
@@ -503,7 +503,7 @@ export class NotScan {
     // nots won't get evaluated during Generic Join since we're never solving for a
     // variable that this scan cares about.
     if((!prejoin || this.args.length)
-       // if we are forcing and not solving for the current variable, then we just accept
+       // if we aren't forcing and not solving for the current variable, then we just accept
        // as it is
        && (!force && !this.internalVars[solvingFor.id] && this.internalVars.length)
        // we also blind accept if we have args that haven't been filled in yet, as we don't
--- a/src/runtime/join.ts
+++ b/src/runtime/join.ts
@@ -71,7 +71,7 @@ export interface ProposalProvider {
   // Check if a prefix of solved variables is a valid potential solution
   // for this provider. SolvingFor is used to ignore accept calls that
   // aren't related to variables the provider is solving for.
-  accept(index: MultiIndex, prefix: any[], solvingFor: Variable, force?: boolean): boolean
+  accept(index: MultiIndex, prefix: any[], solvingFor: Variable, force?: boolean, prejoin?: boolean): boolean
 }
 
 //---------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ibdknox:
This fixes the issue reported on the mailing list [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/eve-talk/Fb82DgqzGVw/EM2oKtokCAAJ).

src/runtime/join.ts
```diff
@@ -497,8 +497,14 @@ export class NotScan {
...
-  accept(multiIndex: MultiIndex, prefix, solvingFor, force?) {
-    if(!force && !this.internalVars[solvingFor.id] && this.internalVars.length || !fullyResolved(this.args, prefix)) return true;
+  accept(multiIndex: MultiIndex, prefix, solvingFor, force?, prejoin?) {
+    // if we're in the prejoin phase and this not has no args, then we need
+    // to evaluate the not to see if we should run. If we didn't do this, arg-less
+    // nots won't get evaluated during Generic Join since we're never solving for a
+    // variable that this scan cares about.
+    if((!prejoin || this.args.length)
+       && (!force && !this.internalVars[solvingFor.id] && this.internalVars.length)
+       || !fullyResolved(this.args, prefix)) return true;
```
joshuafcole: I find this conditional to be pretty hard to parse due to its density. It might be worth commenting about the other early acceptance cases or trying to refactor it to split it up.

This is subjective so I don't know that it's blocking.

src/runtime/join.ts
```diff
@@ -825,13 +831,23 @@ function preJoinAccept(multiIndex: MultiIndex, providers : ProposalProvider[], v
...
           return {accepted: false, presolved};
         }
       }
     }
     ix++;
   }
+  // if we haven't presolved anything, we still need to do a single prejoin pass
+  // to make sure that any nots that may have no external dependencies are
+  // given a chance to end this evaluation
+  if(presolved === 0) {
```
joshuafcole: Even if variables have been presolved, doesn't this pass need to happen? Or is it invoked when those variables are solved? I assumed presolved variables were the set of variables known prior to solving.

src/runtime/join.ts
```diff
@@ -825,13 +831,23 @@ function preJoinAccept(multiIndex: MultiIndex, providers : ProposalProvider[], v
...
+    for(let provider of providers) {
+      if(provider instanceof NotScan && !provider.accept(multiIndex, prefix, {id: "0"}, false, true)) {
```
joshuafcole: what is the shape of `solvingFor` exactly? Why is an object allocation required for each provider?

ibdknox: It's a variable, I can lift the creation.

src/runtime/join.ts
```diff
@@ -497,8 +497,18 @@ export class NotScan {
...
 
-  accept(multiIndex: MultiIndex, prefix, solvingFor, force?) {
-    if(!force && !this.internalVars[solvingFor.id] && this.internalVars.length || !fullyResolved(this.args, prefix)) return true;
+  accept(multiIndex: MultiIndex, prefix, solvingFor, force?, prejoin?) {
+    // if we're in the prejoin phase and this not has no args, then we need
+    // to evaluate the not to see if we should run. If we didn't do this, arg-less
+    // nots won't get evaluated during Generic Join since we're never solving for a
+    // variable that this scan cares about.
+    if((!prejoin || this.args.length)
+       // if we are forcing and not solving for the current variable, then we just accept
```
joshuafcole: if we *aren't* forcing... we just accept as it is, right?

joshuafcole:
:1234: :egg: :egypt:

PR #614, opened 2016-11-30
Fix determining if a stratum has database scans

ibdknox:
Ifs and nots can technically use only local variables and don't necessarily mean we have database scans. The practical implication of this is that any block that had an if or not in it always reran when evaling. This prevents purely static blocks from doing so.


ibdknox:
Other way around. The old code was conservatively saying that anything with an if/not was *definitely* not a one-shot block. Now the code recursively checks inside if/not to see if they're actually static too. If everything is static, including the bodies of the ifs/nots, then we can treat the block as single run.

PR #615, opened 2016-11-30
Special case testing floats to account for rounding/platform differences

cmontella:
The current tests ran into an issue comparing floats, where the generated value was different from the expected value by some rounding error. This was the case that was encountered:

```
⨯ Is tanh[ value: 1 ] returning 0.7615941559557649?
  not ok 249 No matching add found for object: ["tag, div","text, 0.761594155955765"]
    ---
      operator: ok
      expected: true
      actual:   false
      at: resolveActualExpected (C:\Users\Corey\Desktop\Eve\build\test\shared_functions.js:143:24)
```

To fix this, I changed the function that compares sets (`isSetEqual`) to special case float comparisons. If the attribute being compared is called "floatTest", then the comparison will succeed if the expected result and the actual result differ by less than some delta. The delta I chose was arbitrarily `0.00000000001`, and we should discuss exactly what this delta should be.

In implementing this, I faced some choices on how to special case the comparison. I arrived at my solution (using an attribute name as a flag) because `isSetEqual` executes deeply in the call stack, so any extra flags indicating a float comparison would have to be passed through several other functions. We can discuss if there is a more elegant solution here, as I'm not satisfied entirely with it.


test/shared_functions.ts
```diff
@@ -42,10 +41,33 @@ export function eavsToComparables(eavs, entities, index = {}) {
...
   if(as.length !== bs.length) return false;
   for(let a of as) {
-    if(bs.indexOf(a) === -1) return false;
+    let avA = a.split(", ");
+    let attributeA = avA[0];
+    let valueA = avA[1];
+    // If the attribute is called "floatTest" we know to compare
+    // the values to within some delta. Otherwise, we can
+    // do a straight lookup of the key
+    let delta = 0.00000000001;
```
convolvatron: traditionally this is called epsilon :)

test/math.ts
```diff
@@ -1,69 +1,67 @@
...
 
-let constants_list : any = [
-  {"Expression":"pi[]",      "Value":"3.141592653589793"},
-  {"Expression":"e[]",       "Value":"2.718281828459045"},
-  {"Expression":"ln2[]",     "Value":"0.6931471805599453"},
-  {"Expression":"log2e[]",   "Value":"1.4426950408889634"},
-  {"Expression":"log10e[]",  "Value":"0.4342944819032518"},
-  {"Expression":"sqrt1/2[]", "Value":"0.7071067811865476"},
-  {"Expression":"sqrt2[]",   "Value":"1.4142135623730951"}
+let delta = 0.00000000001;
```
convolvatron: thats really the only question about the epsilon(delta), ieee 32 gets you kind of 6-9 and ieee 64 gets you 15-17. just to complicate things intel machines use 80 internally. and truncate to 64 when you are actually saving to a gpr or memory...so...11 is pretty reasonable to assume 64 bit, but if you want to not assume that (and think js asserts that it must be 64), something like 5 would be more appropriate.

this is fine

PR #619, opened 2016-11-30
Blank lookup in a search should produce error

brainopia:
Example
```
search
  lookup[]
```

Will produce error

```
Searching a lookup requires at least one of record, attribute, value, or node to be present
```

_(text may need a proofreading, since I'm not a native English speaker)_

I'll probably send another PR tomorrow to fix `lookup` to work with partially missing variables (related  https://github.com/witheve/Eve/pull/611#issuecomment-263498342).

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/619)

### DIFF
--- a/src/runtime/errors.ts
+++ b/src/runtime/errors.ts
@@ -198,6 +198,12 @@ export function unprovidedVariableGroup(block, variables) {
   return new EveError(id, start, stop, messages.unprovidedVariable(found.name));
 }
 
+export function blankScan(block, scan) {
+  let {id, start: blockStart} = block;
+  let [start, stop] = parser.nodeToBoundaries(scan, blockStart);
+  return new EveError(id, start, stop, messages.blankScan());
+}
+
 export function invalidLookupAction(block, action) {
   let {id, start: blockStart} = block;
   let [start, stop] = parser.nodeToBoundaries(action, blockStart);
@@ -257,6 +263,7 @@ export var messages = {
 
   unimplementedExpression: (op) => `There's no definition for the function ${op}`,
 
+  blankScan: () => 'Lookup requires at least one attribute: record, attribute, value, or node',
   invalidLookupAction: (missing) => `Updating a lookup requires that record, attribute, and value all be provided. Looks like ${missing.join("and")} is missing.`,
 
   neverEqual: (left, right) => `${left} can never equal ${right}`,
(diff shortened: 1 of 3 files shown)


src/runtime/join.ts
```diff
@@ -218,6 +218,109 @@ export class Scan {
     return foo.join("");
   }
 
+  
+  // // I can write macros with eval, hmm
```
cmontella: I think that if this commented code is part of other work, it should be included in a separate PR. For now just take the commented code out of this PR.

test/join.ts
```diff
@@ -1907,6 +1907,22 @@ test("pipe allows you to select ", (assert) => {
...
 })
 
+test("blank lookup errors", (assert) => {
+  let expected = {
+    insert: [],
+    remove: [],
+    errors: true
+  };
+  evaluate(assert, expected, `
+    foo
```
cmontella: No need for this foo

src/runtime/errors.ts
```diff
@@ -257,6 +263,7 @@ export var messages = {
 
   unimplementedExpression: (op) => `There's no definition for the function ${op}`,
 
+  blankScan: () => 'Searching a lookup requires at least one of record, attribute, value, or node to be present',
```
cmontella: I would change the error to:

"lookup requires at least one attribute: record, attribute, value, or node"

test/shared_functions.ts
```diff
@@ -151,6 +151,8 @@ export function evaluate(assert, expected, code, session = new Database()) {
   let {blocks, errors} = builder.buildDoc(parsed.results);
   if(expected.errors) {
     assert.true(parsed.errors.length > 0 || errors.length > 0, "This test is supposed to produce errors");
+  } else {
+    assert.true(parsed.errors.length == 0 && errors.length == 0, "This test is not supposed to produce errors");
```
cmontella: We just merged in changes to shared_functions.ts... this change should merge in fine, but it could cause a conflict. We'll see.

brainopia:
Oops, changes in `join.ts` and `shared_functions.ts` were incidentally in this commit. Seems like when I've decided to re-push commit with amended title, I've accidentally added a few extra things.

All fixed now, thank you!

ibdknox:
Thanks!

PR #620, opened 2016-11-30
Implement client-server interaction via @server and @browser-session

ibdknox:
This adds a new kind of database that attempts to send its adds over the socket to either the client or the server depending on where it's running. It also adds the `--clientAndServer` flag, which allows you to run your program both on the client and server.


ibdknox:
Example program I used to test:

~~~
## Main screen
```
search
  [#active-user name]

bind @browser
  [#div text: "Hello, {{name}}"]
```

## Login screen
Login screen is displayed when there is no active user:
```eve
search
  not([#active-user])

bind @browser
  [#div children:
    [#div children:
      [#label text: "Username:"]
      [#input #user-name type: "text"]]
    [#div children:
      [#label text: "Password:"]
      [#input #user-password type: "password"]]
    [#button #user-submit text: "Login"]]
```

Currently, submitting the form automatically succeeds:
```
search @browser @event
  [#click element: [#button #user-submit]]
  [#user-name value]

commit
  [#active-user name: value]
commit @server
  [#foo name: value]
```

```
search @server
  [#foo name]
commit @browser-session
  [#bar name]
```

```
search @browser-session
  [#bar name]
bind @browser
  [#div text: "hey {{name}} from the server"]
```
~~~

src/runtime/runtimeClient.ts
```diff
@@ -118,6 +118,19 @@ export abstract class RuntimeClient {
         actions.push(new ActionImplementations["+="]("event", e, a, v, "event", scopes));
       }
       this.evaluation.executeActions(actions);
+    } else if(data.type === "result" && (data.database === "server" || data.database === "browser-session")) {
+      // @TODO: should we allow databases other than server at some point?
```
joshuafcole: Yeah, we need to figure out how the extensibility story and policy story interact to give us this if we think it's important. This isn't blocking but it merits a research task.

src/runtime/server.ts
```diff
@@ -159,6 +159,14 @@ class SocketRuntimeClient extends RuntimeClient {
       dbs["editor"] = new BrowserEditorDatabase();
       dbs["inspector"] = new BrowserInspectorDatabase();
     }
+    // in the case where we're running on both the client and the server,
+    // we need to add a browser-session bag and also make our local browser bag
+    // be a normal database that way we don't send UI that is already being handled
+    // by the client
+    if(config.runtimeOwner === Owner.both) {
```
joshuafcole: We could potentially make client-server equipped code run on either/or as well if we always add this and just let it play telephone with itself. I don't know that doing so is particularly useful, but it may be good for testing at least.

ibdknox:
I don't think this is the "right" solution moving forward, so closing this out for now.

Issue #606, opened 2016-11-29
Error messages on fixpoint failure accumulate and push app to the right


convolvatron:
![screen shot 2016-11-28 at 4 02 33 pm](https://cloud.githubusercontent.com/assets/1807982/20691261/9f45ceca-b584-11e6-8a69-eadaebe999dc.png)

PR #622, opened 2016-12-01
Wrap clients injectNotice calls to make it editor-agnostic

joshuafcole:
Reported by @convolvatron, not all calls to `ide.injectNotice` were guarded to ensure that we were running in editor mode. Rather than voiding those messages (as we were in other places) we now shunt them to the console. In the long term we should display them in-browser, but it's an open question as to how to do that without destroying the user-content below. Likely routes include an error modal or an overlay shade.


joshuafcole:
fixes #606

PR #626, opened 2016-12-01
Syntax Diagrams for Eve's grammar.

bd82:
Can be used either as a development tool and/or exposed as part of the docs.


bd82:
Example of Syntax Diagrams (for CSS grammar):
https://rawgit.com/SAP/chevrotain/master/examples/grammars/css/css_diagrams.html

bd82:
You are welcome @ibdknox.

It is interesting to examine complex use cases for the Parsing library I've authored (Chevrotain).
as this is just one more type of feedback which leads to improvements.

PR #630, opened 2016-12-01
Format tests

cmontella:
Just changed some formatting in tests to make more consistent

### DIFF
--- a/test/math.ts
+++ b/test/math.ts
@@ -223,8 +223,7 @@ testSingleExpressionByList(toFixed_list );
 test("Test that string concatenation is still working after NaN change.", (assert) => {
   let expected = {
     insert: [
-      ["a", "tag", "div"],
-      ["a", "text", "Test Testy"],
+      ["a", "result", "Test Testy"],
     ],
     remove: [],
   };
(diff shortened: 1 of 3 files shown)


test/math.ts
```diff
@@ -223,8 +223,7 @@ testSingleExpressionByList(toFixed_list );
 test("Test that string concatenation is still working after NaN change.", (assert) => {
   let expected = {
     insert: [
-      ["a", "tag", "div"],
-      ["a", "text", "Test Testy"],
+      ["a", "result", "Test Testy"],
```
joshuafcole: Not especially important, but what's the rationale for choosing `result` over the existing `text`?

test/math.ts
```diff
@@ -235,8 +234,9 @@ test("Test that string concatenation is still working after NaN change.", (asser
       a = "Test "
       b = "Testy"
       x =  a + b
-    bind @browser
-    [#div text: x]
+
+    bind
+      [result: x]
```
joshuafcole: Do we have an alternative way to see these guys working?

test/math.ts
cmontella: I was making it consistent with the rest of the tests.

PR #639, opened 2016-12-02
implement aggregate join by overriding the whole aggregate() method.

convolvatron:
This version allows for separator to be a set, but doesn't put a separator at the beginning or end of the result string

### DIFF
--- a/examples/join.eve
+++ b/examples/join.eve
@@ -7,8 +7,10 @@
 ```
 
 ```
-   search [#foo name level]
+   search
+       [#foo name level]
        ordinal = sort[value:name given:name]
+
    commit @browser
        [#div text:join[value:name ordinal given:name separator:"/"]]
        [#div text:join[value:name ordinal:0 given:name separator:"/"]]
--- a/examples/join.eve
+++ b/examples/join.eve
@@ -1,0 +1,16 @@
+```
+  commit
+       [#foo name:"a" level:2]
+       [#foo name:"zkp" level:3]
+       [#foo name:"parg" level:0]
+       [#foo name:"naxxo" level:1]
+```
+
+```
+   search [#foo name level]
+       ordinal = sort[value:name given:name]
+   commit @browser
+       [#div text:join[value:name ordinal given:name separator:"/"]]
+       [#div text:join[value:name ordinal:0 given:name separator:"/"]]
+       [#div text:join[value:name ordinal:level given:name separator:"/"]]
+```
--- a/examples/join.eve
+++ b/examples/join.eve
@@ -1,18 +1,18 @@
 ```
   commit
-       [#foo name:"a" level:2]
-       [#foo name:"zkp" level:3]
-       [#foo name:"parg" level:0]
-       [#foo name:"naxxo" level:1]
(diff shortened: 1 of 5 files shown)


examples/join.eve
```diff
@@ -0,0 +1,16 @@
+```
+  commit
+       [#foo name:"a" level:2]
+       [#foo name:"zkp" level:3]
+       [#foo name:"parg" level:0]
+       [#foo name:"naxxo" level:1]
+```
+
+```
+   search [#foo name level]
```
joshuafcole: We should toe the (unfortunately unwritten) style guide in our public examples. @cmontella (we should write that at some point!). We avoid putting patterns on the same line as actions and always include a newline between the previous action's body and the next action. Otherwise this looks :+1:

src/runtime/providers/aggregate.ts
```diff
@@ -176,8 +176,79 @@ export class Max extends Aggregate {
...
+        let s = groups[g].result.sort((a, b) => 
+                                      {if (a.ordinal > b.ordinal) return 1;
+                                       if (a.ordinal === b.ordinal) return 0;
+                                       return -1;})
+        let len = s.length
+        let result = ""
+        for (var i=0; i<len; ++i) {
+            // this means that the sep assocated with a value
+            // is the one which occurs before the value
+            if (i != 0) result += s[i].separator
```
joshuafcole: I thought we decided to only do the interior separators for now? It looks like this guy also attaches a suffix separator.

examples/join.eve
cmontella: http://docs.witheve.com/guides/style/

cmontella:
We had [talked](https://github.com/witheve/rfcs/issues/5#issuecomment-234123785) about the following interface in the past: `text = join(token, index, with)`

I don't know if I like "with" as an attribute as opposed to separator, but I think using token as an argument and having text as the output is good, because it mirrors the split function `(token, index) = split[text, by]`

src/runtime/providers/aggregate.ts
convolvatron: i know i wrote a comment here. we should decide if its the right thing, but the sep (with) precedes  the text (token), so since we lop off the first, only the interstitial are left. i haven't been able to come up with a clever mapping that lets us specify all the options, but but haven't given it too much thought

convolvatron:
corey, i changed the argument names (actually you can use 'value' or 'token' because of some class inheritance rules, but i thought that was ok)

i wrote a basic test, but instead of trying to capture the output database, i wrote another test driver which looks to see if the is a #success in the output, that way the test eve can be self contained and use intermediate data

cmontella:
@convolvatron okay, looks good. I don't think this needs changing in this PR, but playing around with join highlights some issues with aggregates that are unpleasant. For instance, I found I had to do this in order to get the string I wanted: 

```
(token, index) = split[text: "Hello world" by: ""]
text = join[token index given: (token, index) with: "/"]
```

I needed to add the index in to given, or else duplicate letters in the original string went missing. This is an easy thing to overlook. It's also pretty verbose for just a string join, but that's what it takes.

PR #641, opened 2016-12-04
Fix a bug in "gaussian" function

pithub:
The Gaussian provider uses different keys when reading and writing into the internal cache, so it violates referential transparency as describe here: https://gist.github.com/joshuafcole/52f706e0e838b3fb7536814de8879af0.


cmontella:
Thanks for spotting this!

PR #644, opened 2016-12-04
Added length to strings

jimmyhmiller:
I added length. Everything seems to be working. Not sure if the tests I've added are sufficient or if you'd like some more/different ones. Let me know if you want anything changed.


cmontella:
Thanks for this! Before we merge in, we need to consider the different ways to measure the length of a string. For instance, what should be the length of a string that contains characters that span more than one byte? For instance, `length[text: "𝐁"] = 2` in the current implementation.

I think this function should take an additional parameter to specify how the length of the string is determined. We can call it "as" and it can take the values "symbol", "code-point", and "byte" with a default to "symbol".

In this case

```
length[text: "𝐁"] = 1
length[text: "𝐁" as: "byte"] = 2
```

I think this is less surprising in the common case, and anyone who wants to deal with byte lengths specifically can choose to do so.

Thoughts?

src/runtime/providers/string.ts
```diff
@@ -251,7 +251,37 @@ class Urlencode extends Constraint {
...
 }
 
+class Length extends Constraint {
+  static AttributeMapping = {
+    "text": 0,
+  }
+  resolveProposal(proposal, prefix) {
+    let {args} = this.resolve(prefix);
+    let [text] = args;
+    return [text.length];
```
cmontella: Return the length of the string in terms of symbols

cmontella:
@jimmyhmiller Just wanted to check if you wanted to implement these changes or if  you wouldn't mind us picking it up?

jimmyhmiller:
I can pick it up. Sorry about the delay.

cmontella:
No worries! The only reason I asked was I didn't want to step on your toes, if you already had started the changes.

jimmyhmiller:
Wasn't 100% sure how you wanted to deal with the case of "as" that was wrong. Feel free to make any changes you need. I probably wouldn't be able to get back right away. Though I'm hoping to make more PR's soon with more string functions. I [had mentioned](https://github.com/witheve/rfcs/issues/5#issuecomment-264734463) a higher-order way of doing that. Was hoping to get your feedback on that sort of approach before I did. 

(Another option would just be some helper functions to make it easier to implement these constraints in the common case.)

cmontella:
Okay, I fixed a couple issues, but the changes are in a new PR here: https://github.com/witheve/Eve/pull/700

Thanks for your help on this!

PR #657, opened 2016-12-06
Fix stratification when joining aggregate results with other scans

### LINKED ISSUE #649: Aggregate stratifier renders some queries unpleasant
When authoring a unit testing framework example I ran into an obnoxious query. Filtering by the result of an aggregate by any value naturally joined with one of it's inputs is impossible to do in the same block. The stratifier deems this to be a cycle since the aggregate's output may filter its inputs. The obvious intended result in this example is for the aggregate to run on its inputs and post-filter.

This isn't a show-stopping issue because it can be rewritten into two blocks, but it is likely to surprise any users who run across this.

Commit a test case to run.

``` eve
commit @test
  [#test input:
    [#add a: 7, b: 3]
    [#add a: 6, b: 4]
    [#add a: 1, b: 1]
   expected:
     [#result value: 10]
     [#result value: 2]]
```

Kick off the test case.

``` eve
search @test
  [#test input]
  lookup[record: input, attribute, value]

bind
  lookup[record: input, attribute, value]
```

Derive some metadata about the test cases.

``` eve
search @test
  test = [#test expected]
  lookup[record: expected, attribute, value]
  av-count = count[given: (attribute, value), per: expected]
  test-count = count[given: expected, per: test]

bind @test
  expected.av-count := av-count
  test.count := test-count
```

Confirm the test passes.

``` eve
search @test
  test = [#test expected]
  av-count = expected.av-count
  test-count = test.count
search
  not(actual = [tag: expected.tag]
               lookup[record: expected, attribute, value]
               lookup[record: actual, attribute, value]
               av-count != count[given: (attribute, value), per: actual])
  test-count = count[given: expected, per: test] // The problem is here

bind
  test.passed := true
```

The block to test.

``` eve
search
  [#add a b]

bind
  [#result value: a + b]
```

ibdknox:
The issue in #649 was caused by a set of issues. First, the maybeLevel function had bugs in it that caused us to look at the wrong scan to determine the minimum level for a variable. Second, we were automatically setting the returns of an aggregate to the same level as the aggregate itself, which is incorrect in the case where there are other providers for a variable (e.g. a database scan).

To fix this, I fixed the bugs in maybeLevel and then instead of blanket setting all returns of aggs to the aggs' level, I do a pass to check for the minimum level that could provide that var and set it to that.

Fixes #649.

### DIFF
--- a/src/runtime/builder.ts
+++ b/src/runtime/builder.ts
@@ -558,6 +558,16 @@ function buildActions(block, context, actions, scans) {
 // Stratifier
 //-----------------------------------------------------------
 
+function isAggregate(scan) {
+  if(scan instanceof Aggregate ||
+     scan instanceof Sort ||
+     scan.hasAggregate ||
+     (scan.strata && scan.strata.length > 1)) {
+    return true;
+  }
+  return false;
+}
+
 function stratify(scans) {
   if(!scans.length) return [new BlockStratum([], [])];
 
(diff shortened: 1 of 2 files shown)


joshuafcole:
One example this may break is:

``` eve

[#foo a]
[#bar b]

a = count[given: b]
final = sum[value: a]

```

The count aggregate should be on level 1 and the sum aggregate should be on level 2 to match expectations, but using the maximum min level of all input variables will put them both on level 1.

PR #665, opened 2016-12-07
Add Fix function.

Cormac-Williams:
Mentioned as missing in https://github.com/witheve/docs/issues/46

### DIFF
--- a/src/runtime/providers/math.ts
+++ b/src/runtime/providers/math.ts
@@ -328,6 +328,13 @@ class Round extends ValueOnlyConstraint {
   }
 }
 
+class Fix extends ValueOnlyConstraint {
+  getReturnValue(args) {
+    let x = args[0];
+    return x - x % 1;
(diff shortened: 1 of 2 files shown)


cmontella:
Thanks for this! The reason I wrote the doc in the first place was because I was copying off of some matlab docs, and that's the one I happened to have opened. Hopefully this is just what someone was looking for :)

Cormac-Williams:
You're welcome. I should have noted that this is a re-implementation of the Trunc function from Javascripts Math. I didn't catch cos I thought it was already done. So I have kept the naming in line with the docs. It may be a confusion to newcomers who recognise the naming of all the other functions to keep the name as is, but obviously it's more work to change both. Let me know if you guys want to re-name and I'll help out.

Regards,
Cormac.

cmontella:
Looks good, thanks, we'll keep it as is for now.

Issue #670, opened 2016-12-08
Concat Error investigation leads to bug in Builder.ts


Cormac-Williams:
If you have an function providor defined without an Attribute Mapping, such as concat[ any args ] and then use it with args like so

search
s1 = "Test"
s2 = "Testy"
x = concat[a:s1 b:s2]

You get weird syntax errors in the editor and an error in the console
builder.ts:407 Uncaught TypeError: Cannot read property 'a' of undefined

Fix pull will follow shortly.
Regards,
Cormac.

PR #671, opened 2016-12-08
Fix for 670, handle undefined attribute mapping.

Cormac-Williams:
For https://github.com/witheve/Eve/issues/670

### DIFF
--- a/src/runtime/builder.ts
+++ b/src/runtime/builder.ts
@@ -424,14 +424,15 @@ function buildExpressions(block, context, expressions, outputScans) {
         return;
       }
       for(let attribute of expression.record.attributes) {
-        let ix = impl.AttributeMapping[attribute.attribute];
-        if(ix !== undefined) {
+        let ix; 
+        if(impl.AttributeMapping && (ix = impl.AttributeMapping[attribute.attribute]) !== undefined) {
           args[ix] = context.getValue(attribute.value);
         } else if(impl.ReturnMapping && (ix = impl.ReturnMapping[attribute.attribute]) !== undefined) {
           results[ix] = attribute.value;
         } else {
           // @TODO: error - unknown arg/return for the function call
         }
+
       }
       let resultIx = 0;
       for(let result of results) {
--- a/src/runtime/builder.ts
+++ b/src/runtime/builder.ts
@@ -430,7 +430,7 @@ function buildExpressions(block, context, expressions, outputScans) {
         } else if(impl.ReturnMapping && (ix = impl.ReturnMapping[attribute.attribute]) !== undefined) {
           results[ix] = attribute.value;
         } else {
-          // @TODO: error - unknown arg/return for the function call
+          context.errors.push(errors.unrecognisedFunctionAttribute(block , expression,attribute));
         }
 
       }
--- a/src/runtime/errors.ts
+++ b/src/runtime/errors.ts
@@ -240,6 +240,11 @@ export function incompatabileTransitiveEquality(block, variable, value) {
   return new EveError(id, start, stop, messages.variableNeverEqual(variable, variable.constant, value));
 }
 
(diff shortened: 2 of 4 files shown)


cmontella:
One question I have is what should happen in this case. For example, `pi[]` suffers from the same problem, but this fix will cause e.g. `pi[a: "foo"]` evaluate as normal. Should this be an error? I'm leaning toward yes. Thoughts?

Cormac-Williams:
As far as I can see attributes in Pi[] are meaningless, but if put in should not cause an error at the console level or blow up the expression builder, same with concat. 

I noticed because the code for Concat is written so that all the attributes are Joined, with no respect to the name of the attribute, but if there is no attribute mapping assigned at all, you get this error. This is quite a flexible pattern to retain for Function writers, but apart from Concat I cannot think of any particular use cases. 

In fairness, I don't know whether Concat is planned to be replaced by Join. I was checking that you could still add strings easily once + stops handling strings, you can do it explicitly with string interpolation, so Concat is in some ways extraneous.

Let me know which way you guys want to go.
Regards,
Cormac.

cmontella:
`concat[]` is the engine behind string interpolation. e.g. "{{name}} {{age}}" is the same as `concat[a: name b: " " c: age]`. We're not planning on exposing `concat[]` in the same way as the other functions in favor of string interpolation.

With something like `pi[]`, if we think back to what a function is under the covers, it would be `[#fn #pi value]`, where value is the constant `3.1415...`. If we add an argument to this record `[#fn #pi value a]` it should return nothing, since there is no relation that matches this attribute pattern.

cmontella:
Just wanted to update this PR a little... since this was proposed, we changed concat's naming to make it more explicit that it is for internal use only:  https://github.com/witheve/Eve/blob/master/src/runtime/providers/string.ts#L349

We discussed my original question (pi[a: "foo"] evaluate as normal. Should this be an error? ) in the office, and came to the conclusion that for things like constants, `pi[a: "foo"]` should return nothing, because there's no matching relation.

Do you want to make this change? Let me know otherwise I'll make the change and merge it in. Thanks for your help on this!

Corey

Cormac-Williams:
I'll have a look at this change tomorrow.
Regards,
Cormac.

Cormac-Williams:
Sorry for the slow progress here. I've started this by adding a ide error for an unrecognised attribute. To make clear which attributes are causing any filtering.

Cormac-Williams:
Ok I going to pause on this for now and seek guidance.

The addition of the context.errors.push may have handled 90% of the issue. By raising an error the block no longer executes,  so this code no longer returns a result

~~~
    search
      x = pi[invalid-attribute:0]
    bind
      [PI: x]
~~~

and you cannot use the internal concat function publicly but interpolation still works. So

~~~
search
  a = "11"
  b = "22"
  x =  eve-internal/concat[a b]
commit @view
  [#value value:"{{x}}"]
~~~

produces errors in the browser and does not function.

Only problem is that code such as this

    ~~~
    search
      x = if pi[invalid-attribute:0] then
             1
          else
             0
    bind
      [PI: x]
    ~~~

does the same.

I cannot think of any occasion when you would want to write this type of code, but may be wrong.

I was assuming that I'd have to start changing various Eve functions / constraints to produce this change so let me know if there is further work to be done here. Otherwise I'll put together some tests to finish this pull off.

Regards,
Cormac.

cmontella:
Thanks for getting back to me on this Cormac, In terms of using the internal functions, I think users shouldn't be exposed to them, as so them breaking isn't an issue (as long as string interpolation still works, as you noted). I think this is fine, so we can merge once the tests are added.

Cormac-Williams:
Done

Regards,
Cormac.

cmontella:
Thanks again!

Issue #673, opened 2016-12-09
Using parens around sub-records prevents the parent from being part of the identity


ibdknox:
The following blocks should be identical, but don't behave identically.

This works:
```
search
   menu-item = [#menu name image cost]
bind @browser
   [#div menu-item #item-container
     children:
      [#div text:name]
      [#div text:"${{cost}}"] ]
```

This doesn't:
```
search
   menu-item = [#menu name image cost]
bind @browser
   [#div menu-item #item-container
     children: ([#div text:name]
                      [#div text:"${{cost}}"] )]
```

The suspected reason is that we aren't sending the parent down through the parenthesis as a dependency for the sub-records.

PR #674, opened 2016-12-09
Make sure records in parenthesis are treated the same

ibdknox:
In code where sub-records were defined in a parenthesis, we weren't pushing the parent's identity down into the children, we also weren't setting `eve-auto-index` on them. This makes sure we do both and should remove any differences between using parens or not.

Fixes. #673

### DIFF
--- a/src/runtime/parser.ts
+++ b/src/runtime/parser.ts
@@ -882,6 +882,31 @@ export class Parser extends chev.Parser {
       self.OR2([
         {ALT: () => {
           result = self.SUBRULE(self.infix);
+          // if the result is a parenthesis, we have to make sure that if there are sub-records
+          // inside that they get eve-auto-index set on them and they also have the parent transfered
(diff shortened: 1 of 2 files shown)


ibdknox:
`eve-auto-index` serves a pretty important pragmatic purpose, but maybe there's some other way to accomplish what it gives us. The main value is in something like this:

```
bind @browser
  [#div children: 
     [#div text: "foo"]
     [#div text: "bar"]
     [#div text: "baz"]]
```

Since `eve-auto-index` is applied to these, we know to keep them in that order on display. If we didn't have something like that you'd need to supply `sort` on all of them despite the fact that we've already expressed the intent.

ibdknox:
Where that overhead would really get you is if you were inserting something in the middle and then had to update every `sort` attribute underneath. Do that a couple times and it'll seem worse than busy work.

PR #675, opened 2016-12-09
Feature/save load gist

joshuafcole:
Enables

* Saving current document as markdown to gist
* Loading documents from gists in-editor
* Loading documents from gist-formatted file ids in the editor.

Blocking tasks:

- [x] Silently remap local anchor tag hrefs to ignore the IDE portion of the document fragment. This isn't any more broken than it used to be, but the problem becomes more prevalent since the documentId is no longer a known static value.
- [x] Handle saving edits to documents loaded from gists in some reasonable way.

Notes:

- Code quality is lacking here. Improvements are blocked by some refactoring work that isn't critical. it's still worth discussing any specific issues found, but it's important that we revisit this in the near future for cleanup as part of the broader file store rework.

### DIFF
--- a/src/ide.ts
+++ b/src/ide.ts
@@ -94,8 +94,10 @@ class Navigator {
 
     let root:TreeNode = this.nodes[id];
     if(!root) {
-      console.error("Cannot load non-existent document.");
-      return;
+      //console.error("Cannot load non-existent document.");
+      //return;
+
+      root = this.nodes[id] = {id, type: "document", name};
     }
     root.open = true;
     root.children = undefined;
@@ -407,8 +409,8 @@ class Navigator {
     let type = this.currentType();
     return {c: "navigator-header", children: [
       {c: "controls", children: [
-        this.open ? {c: `up-btn flex-row`, click: this.navigate, children: [
-          type === "document" ? {c:  `up-btn ion-android-arrow-up ${(type === "folder") ? "disabled" : ""}`} : undefined,
+        this.open ? {c: `up-btn flex-row  ${(this.currentId === this.rootId) ? "disabled" : ""}`, click: this.navigate, children: [
+          {c:  "up-icon ion-android-arrow-up"},
           {c: "label", text: "examples"}
         ]} : undefined,
         {c: "flex-spacer"},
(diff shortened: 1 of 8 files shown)


ibdknox:
Is this able to load regular github links yet?

ibdknox:
Seems to work fine for gists :) I think we probably want to make some changes here to make it a little more obvious UX-wise:

- [ ] We should get rid of my on hover for hiding for the nav so that it's always there
- [ ] The load button at least should show up on the directory part of the nav
- [ ] The saved to gist notification should probably be a bit more noticable
- [ ] It'd be nice to have some feedback when you initially press the save button that lets you know it's doing something
- [ ] Collapsing the nav now puts the arrow in a strange place

src/ide.ts
```diff
@@ -91,8 +94,10 @@ class Navigator {
 
     let root:TreeNode = this.nodes[id];
     if(!root) {
-      console.error("Cannot load non-existent document.");
-      return;
+      //console.error("Cannot load non-existent document.");
+      //return;
+
+      root = this.nodes[id] = {id, type: "document", name};
```
ibdknox: Is this always right? What happens if I do try to load something non-existent locally?

src/ide.ts
```diff
@@ -387,23 +417,35 @@ class Navigator {
...
-          {c: "label", text: "examples"},
+        this.open ? {c: `up-btn flex-row  ${(this.currentId === this.rootId) ? "disabled" : ""}`, click: this.navigate, children: [
+          {c:  "up-icon ion-android-arrow-up"},
+          {c: "label", text: "examples"}
         ]} : undefined,
         {c: "flex-spacer"},
-        {c: `${this.open ? "expand-btn" : "collapse-btn"} ion-ios-arrow-back`, click: this.togglePane},
+
+        this.open && type === "document" ? {c: "ion-ios-cloud-upload-outline btn", title: "Save to Gist", click: () => this.ide.saveToGist()} : undefined,
+        this.open && type === "document" ? {c: "ion-ios-cloud-download-outline btn", title: "Load from Gist", click: this.openLoadDialog} : undefined,
```
ibdknox: seems like I should be able to load a gist regardless of where I am

joshuafcole:
@ibdknox thoughts on making the saved notice more visible without being obnoxious?

ibdknox:
Nav is empty when:
1) Load, change, load, change, click overwrite

On change the new file doesn't show up in the directory index without a refresh.

joshuafcole:
This is ready to ship pending final approval.

PR #687, opened 2016-12-12
don't allow bad responses to flow through the http code for node

### LINKED ISSUE #668: Bad @http code causes the server to crash.
Playing with the code from another issue I found bad @http code added to quickstart.eve causes the server to crash.

If you add this to the quickstart.eve

commit @http
   [#request #get-all url: "http://localhost:8081/getAll"]

Stop and start the server, possibly do npm install again (I was playing with a couple of things.)
Make the request to localhost:8080.

Server stops with the message

/home/noiwont/src/Eve/build/src/runtime/server.js:280
            throw err;
                  ^
TypeError: Cannot read property 'headers' of undefined
    at Request._callback (/home/noiwont/src/Eve/build/src/runtime/databases/node/http.js:50:25)
    at self.callback (/home/noiwont/src/Eve/node_modules/request/request.js:186:22)
    at Request.EventEmitter.emit (events.js:95:17)
    at Request.onRequestError (/home/noiwont/src/Eve/node_modules/request/request.js:845:8)
    at ClientRequest.EventEmitter.emit (events.js:95:17)
    at Socket.socketErrorListener (http.js:1547:9)
    at Socket.EventEmitter.emit (events.js:95:17)
    at net.js:441:14
    at process._tickCallback (node.js:415:13)

I wasn't running anything on 8081 and was expecting the record to reflect the issue.

Also seems to be running in the node version of @http, so I'm assuming the code got compiled in and then runs on the server side rather than client side.

Regards,
Cormac.

ibdknox:
Fixes #668

### DIFF
--- a/src/runtime/databases/node/http.ts
+++ b/src/runtime/databases/node/http.ts
@@ -32,6 +32,7 @@ export class HttpDatabase extends Database {
     httpRequest(options, (error, response, body) => {
       if(error || !response) {
         // @TODO: expose errors and other weirdness into Eve instead of just bailing here.
+        console.error(error || "ERROR: No response received from HTTP request");
         return;
       }
       let scope = "http";
--- a/src/runtime/databases/node/http.ts
+++ b/src/runtime/databases/node/http.ts
@@ -30,10 +30,10 @@ export class HttpDatabase extends Database {
       options.body = request.body[0];
     }
     httpRequest(options, (error, response, body) => {
-      // console.log("GOT RESPONSE", response.statusCode);
-      // console.log(error);
-      // console.log(response);
-      // console.log(body);
+      if(error || !response) {
+        // @TODO: expose errors and other weirdness into Eve instead of just bailing here.
+        return;
+      }
       let scope = "http";
       let responseId = `${requestId}|response`;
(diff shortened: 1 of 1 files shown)


ibdknox:
@convolvatron this is the band-aid to fix #668, but I've left a todo in there for you to add whatever you were thinking for handling error responses and such. One thing to keep in mind is that whatever we do here we also need to do in the browser-based http database as well.

src/runtime/databases/node/http.ts
```diff
@@ -30,10 +30,10 @@ export class HttpDatabase extends Database {
       options.body = request.body[0];
     }
     httpRequest(options, (error, response, body) => {
-      // console.log("GOT RESPONSE", response.statusCode);
-      // console.log(error);
-      // console.log(response);
-      // console.log(body);
+      if(error || !response) {
+        // @TODO: expose errors and other weirdness into Eve instead of just bailing here.
```
joshuafcole: Is there any reason not to stick a `console.warning(...)` here to indicate the failure in the interim?

ibdknox: fixed

Issue #624, opened 2016-12-01
It is quite difficult to specify a "default only" value for checked and my attempts result in multiple checked values


frankier:
For example, see the settings screen of Minesweeper. Here: https://github.com/frankier/eve-experiments/blob/22dc4b99ea269ee392b1accd58eb5e84c5ceaca2/minesweeper.eve#L436 I maintain a #difficulty-changed record as a flag to see if the user has interacted with the checkboxes. I bind the default value only if not(#difficulty-changed). This is clearly a bit awkward. Perhaps there could be a React like defaultChecked (probably default-checked in our case) attribute.

I guess this is a separate issue, but to run Minesweeper successfully even this doesn't work since I get the error message: Unable to set 'checked' multiple times on entity. It works fine if I just disable the check as I've done here: https://github.com/frankier/Eve/commit/91ff96acdc5a0a24285b2098e6ddf1e223900630

frankier:
I've started working on the first bit of this.

joshuafcole:
hey @frankier,

The current solution to default values in inputs is through binds. It looks like this:

``` eve
search @browser
  value = if [#my-input value] then value
               else "some default"

bind @browser
  [#my-input value]
```

Of course, you can use ifs for defaults anywhere in the stack, not just for DOM. We're keeping an eye on if this proves confusing or unpleasant for users though. If so we'll consider including `defaultValue` attributes ala React.

If you have any other questions or comments about this, feel free to reopen this issue.

PR #691, opened 2016-12-13
Make client/browser checks for multiton attribute values use set semantics

frankier:
Following on from https://github.com/witheve/Eve/issues/624

Now I know the correct way to deal with default values, I can file a better bug report and this PR. Given the following example:

```
commit @browser
  [#label text: "I want to be asked a question:" children: [#input type: "checkbox" name: "panel" value: "show"]]
```
```
search @browser
  input = if [#input name: "color" checked: true] then [#input name: "color" checked: true]
          else [#input name: "color" value: "other"]
bind @browser
  input.checked := true
```
```
search @browser
  [#input name: "panel" value: "show" checked: true]
bind @browser
  [#h3 text: "What's your favorite color?"]
  [#label text: "Red" children: [#input type: "radio" name: "color" value: "red"]]
  [#label text: "Blue" children: [#input type: "radio" name: "color" value: "blue"]]
  [#label text: "Some other color" children: [#input type: "radio" name: "color" value: "other"]]
```

Steps to reproduce:

1. Click "I want to be asked a question"
2. Click "Red"
3. Click "I want to be asked a question" twice

Actual behaviour:
See the error: "Unable to set 'checked' multiple times on entity" in the console.
Note that now nothing is selected.

Expected behaviour:
No error.
The default is still selected.

Proposed fix in this PR: since it appears that although Eve attribute values behave as sets, they can end up with multiple identical values, all the checks of whether they are multiton sets must be changed from checking if there is more than one array element to check if there is more than one distinct value. The other possible fix would be to make sure it was impossible for the attribute values to have non-distinct values, but some shallow investigation led me to believe the current situation of "multisets that get interpreted as sets" is by-design.

P.S. Here is a snippet I used for tracing what was going on in my example:
```
search @browser
  input = [#input name value]
  checked = if input.checked then "✔"
            else "✘"
bind @browser
  [#div text: "Current status" children:
    [#div text: "{{name}} {{value}} {{checked}}"]]
```
```
search @event @session @browser
  [#change element: [#input name value] checked]
  [#time frames]
commit @browser
  [#div sort: frames text: "{{frames}}: {{name}} {{value}} => {{checked}}"]
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -83,7 +83,13 @@ function handleDiff(state, diff) {
     dirty.insert(e, a);
 
     if(!entity[a]) entity[a] = [];
-    entity[a].push(v);
+    let isSingleValueAttr = a == "value" ||
+                            a == "tag" ||
+                            a == "sort" ||
+                            a == "checked";
+    if(!(isSingleValueAttr && entity[a].length == 1 && entity[a][0] === v)) {
+      entity[a].push(v);
+    }
 
     // Update indexes
     if(a === "tag") indexes.byTag.insert(v, e);


joshuafcole:
Hey @frankier, sorry for the long wait. Thanks for pitching in on this issue. I think it might make a little more sense to just de-dupe the facts when they arrive in the client around here: https://github.com/witheve/Eve/blob/master/src/client.ts#L35 That way we don't need to do extra work in all the individual branches.

frankier:
Ah yes that's much better!

joshuafcole:
Hey @frankier, sorry for not responding earlier, we've forward-ported your suggestions to 0.3 and 0.4, which will be released in the next few weeks. Since this patch has several merge conflicts which would need to be resolved (and retargeted against the #release branch) and 0.2 is rapidly approaching end of life, I'm going to close this PR. If you feel strongly that you still want to get this into 0.2 before we retire it, ping me and I'll be happy to help you out. Otherwise, you should find things all patched up in 0.4. :)

PR #696, opened 2016-12-13
Initdb - preload server with records from json

convolvatron:
this patch creates a -json flag which reads one or more passed json files, and stores the objects in @init 

if the file is preceded by a colon separated field foo:file, then "foo" is created as the target database

if the database is preceded by a colon separated tag, then each json object in file will have #tag added to its record (for tables dumped as sets of objects)

### DIFF
--- a/bin/eve.js
+++ b/bin/eve.js
@@ -54,7 +54,7 @@ if(argv["help"]) {
     --json          Load the passed json file into @init. One additional
                     colon separated argument to specify the target database,
                     an additional to add the passed tag to each json record
-                    [[tag:]database]:filename
+                    filename[:database[:tag]]
 
     If the Eve binary is run in a project directory (a directory containing a
     package.json file), it will use that directory as your workspace. Otherwise
--- a/bin/eve.js
+++ b/bin/eve.js
@@ -104,6 +104,8 @@ if(!filepath) {
 let mode = Mode.workspace;
 if(filepath && !editor) mode = Mode.file
 
+if (!Array.isArray(initJsonDB) && initJsonDB) initJsonDB = [initJsonDB];
+
 var opts = {internal: internal, runtimeOwner: runtimeOwner, controlOwner: controlOwner, editor: editor, port: port, path: filepath, internal: internal, root: root, eveRoot: eveRoot, mode, initJsonDB:initJsonDB};
 config.init(opts);
 
--- a/bin/eve.js
+++ b/bin/eve.js
@@ -10,7 +10,8 @@ var Owner = config.Owner;
 var Mode = config.Mode;
 var server = require("../build/src/runtime/server");
 
-const argv = minimist(process.argv.slice(2), {boolean: ["help", "version", "localControl", "server", "editor"]});
+const argv = minimist(process.argv.slice(2), {boolean: ["help", "version", "localControl", "server", "editor"],
+                                              string:["json"]});
 
 // Since our current development pattern uses npm as its package repository, we treat the nearest ancestor directory with a package.json (inclusive) as the directory's "root".
 function findRoot(root) {
@@ -32,7 +33,7 @@ var controlOwner = argv["localControl"] ? Owner.client : Owner.server;
 var editor = argv["editor"] || false;
 var filepath = argv["_"][0];
 var internal = false;
-
+var initJsonDB = argv["json"]
 var root = findRoot(process.cwd());
 var eveRoot = findRoot(__dirname);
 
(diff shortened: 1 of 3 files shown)


convolvatron:
tag mode hasn't been tested yet, but database specification and multiple files have been

just looking for initial feedback, there will probably be a fix for tags arriving

bin/eve.js
```diff
@@ -50,6 +51,10 @@ if(argv["help"]) {
     --port <number> Change the port the Eve server listens to (default 8080).
     --localControl  Entirely disable server interaction. File changes will be
                     stored in localStorage.
+    --json          Load the passed json file into @init. One additional
+                    colon separated argument to specify the target database,
+                    an additional to add the passed tag to each json record
+                    [[tag:]database]:filename
```
joshuafcole: Might be nice to invert the order here so the optionals come last in the order they'd get added. I don't feel super strongly on this. 3/10 maybe.

src/config.ts
```diff
@@ -10,7 +10,8 @@ export interface Config {
   root?: string,
   eveRoot?: string,
   internal?: boolean,
-  mode?: Mode
+  mode?: Mode,
+  initJsonDB? : any
```
joshuafcole: If we can give a specific type to this we should. Is there a reason it needs to be `any`?

src/runtime/server.ts
```diff
@@ -44,15 +47,18 @@ global["browser"] = false;
...
+  constructor(databases) {
     let server = new ServerDatabase();
     const dbs = {
       "http": new HttpDatabase(),
       "server": server,
       "shared": shared,
       "browser": new Database(),
     }
-    super(dbs);
+    var uniondb = {}
```
joshuafcole: These `var`s should be `let`s in TS land. `let` gives us sane scoping rules. You really don't want to think about the shadowing BS var does.

src/runtime/server.ts
```diff
@@ -295,12 +336,17 @@ export function run() {
     }
   }
 
-  let app = createExpressApp();
+  var databases;
+  if (config.initJsonDB) {
+    databases = loadDatabases(config.initJsonDB);
+  }
+
+  let app = createExpressApp(databases);
```
joshuafcole: Rather than plumbing the list of databases throughout, let's attach it to the config. That way we can access it directly where it's needed. In the future if we want to have separate instances running in the same process for some reason we can give each instance its own config object.

joshuafcole:
A few tiny cleanup things, and one suggested refactor. In general looks pretty solid.

src/runtime/server.ts
```diff
@@ -269,6 +278,38 @@ function initWebsocket(wss, withIDE:boolean) {
...
+
+  for (var i = 0; i < arrayLength; i++) {
+    var database = "init";
+    var prefix = undefined;
+    var terms = dbs[i].split(":")
+    var count = 0;
+
+    if (terms.length == 3) prefix = terms[count++];
+    if (terms.length == 2) database = terms[count++];
+    var data = fs.readFileSync(terms[count]).toString();
```
joshuafcole: Oh, and we need to wrap this in `if(fs.existsSync(<path>) { ... } else { /* handle error */ }` or we'll explode if the file is missing.

bin/eve.js
convolvatron: well, it clearly doesn't matter semantically. i guess i was thinking kinda like the database is the containing object? the tag is kinda the same? i'll flip it unless you think we should get a third opinion

src/runtime/server.ts
convolvatron: had no idea, i'll scrub my vars...why did B.E. make such a bad scoping...i guess it was supposed to be for lua-like use cases

src/runtime/server.ts
convolvatron: but..dont we want it to?

convolvatron:
tag mode has been tested...there was a comment concerning i/o error handling that got lost? i suggested we didn't really need to catch the error and throw our own

jpka:
Trying to test this branch as I am interested in the feature but I can't seem to get it to work. What is the structure of the json suposed to be? It doesn't throw any errors, just doesn't load the data. Not to the default @init db, not to a custom db, and not with the tag param either

joshuafcole:
Juan:

You'll need to run it with the --server flag, since the client can't access
the file system and we haven't added a pass-through for this feature yet.

jpka:
@joshuafcole  Oh I see. Thanks for the clarification.

ibdknox:
Closing for now.

PR #700, opened 2016-12-14
Add length[] function

cmontella:


cmontella:
An extension of https://github.com/witheve/Eve/pull/644

Credit goes to @jimmyhmiller for this one!

cmontella:
A follow-on to the PR is the documentation for length[]: https://github.com/witheve/docs/pull/60

src/runtime/providers/string.ts
```diff
@@ -251,7 +251,83 @@ class Urlencode extends Constraint {
...
+  // Adapted from: https://mathiasbynens.be/notes/javascript-unicode
+  countSymbols(string) {
+    let index;
+    let symbolCount = 0;
+    for (index = 0; index < string.length - 1; ++index) {
+      var charCode = string.charCodeAt(index);
+      if (charCode >= 0xD800 && charCode <= 0xDBFF) {
+        charCode = string.charCodeAt(index + 1);
+        if (charCode >= 0xDC00 && charCode <= 0xDFFF) {
+          ++index;
```
joshuafcole: Super unimportant, but let's avoid using the pre-increment for clarity reasons.

src/runtime/providers/string.ts
```diff
@@ -251,7 +251,83 @@ class Urlencode extends Constraint {
...
+    } else if (az === "code-points") {
+      return [text.length];
+    }
+    return undefined;
+  }
+
+  resolveProposal(proposal, prefix) {
+    let {args} = this.resolve(prefix);
+    let [text, az] = args;
+    if (az === undefined) {az = "symbols"};
```
joshuafcole: Let's either drop the braces or indent this appropriately. This looks confusingly like an object.

test/strings.ts
```diff
@@ -25,3 +25,154 @@ test("test string join ordering", (assert) => {
   `);
   assert.end();
 });
+
+test("check length of hello", (assert) => {
```
joshuafcole: These may need touched depending on whether they go in before the async meat-grinder work Chris has going on.

test/strings.ts
```diff
@@ -25,3 +25,154 @@ test("test string join ordering", (assert) => {
...
+      len = length[text: "hello" as: "asdfasdf"]
+
+    commit
+      [len: len]
+    ~~~
+  `);
+  assert.end();
+})
+
+test("check length multi-byte characters as symbols", (assert) => {
```
joshuafcole: We should add a test-case where symbols and code-points would differ.

test/strings.ts
cmontella: What do you mean by "touched" ?

joshuafcole:
✈️  🛩  ☁️

PR #707, opened 2016-12-14
mark a few providers as being internal

ibdknox:
There are a few functions that aren't meant for end-users that the runtime itself uses. This PR renames those and adds comments to make that clearer so people don't waste their time trying to get them to work.


cmontella:
Looks good!

ibdknox:
For observers, this diff looks insane because git decided that `And` and `Or` were actually the same thing and so all the falses and trues get swapped (I removed Or).

PR #715, opened 2016-12-16
Lookup with missing variables

### LINKED ISSUE #685: Lookup with missing attributes
Hello,

As I've previously mentioned in https://github.com/witheve/Eve/pull/611#issuecomment-263498342, currently `lookup` works only if record-attribute-value or record-attribute-value-node specified. If any of record/attribute/value is missing then `lookup` will break.

There are multiple ways to fix it.

For example, substitute missing parts with new variables https://github.com/brainopia/Eve/commit/0d8209a773bac13272bd8749362bf4082e92aa09. But then unnecessary variables will be solved for.

Or introduce hidden variables  https://github.com/brainopia/Eve/commit/c32bbd10671eee209fc5e76af907907fb1784a16 then only intermediate hidden variables that are needed will be solved for. But there is still overhead from extra round solving wasted on intermediate variables instead of solving directly appropriate variables at once.

A faster approach would look like https://github.com/brainopia/Eve/commit/692024d6144a23f7e3c22fc7554fbb44e03b4e7d  but since it relies on full-scan when missing variable is solved for, it'd need a full-scan optimization like @ibdknox mentioned at https://github.com/witheve/Eve/pull/611#pullrequestreview-11040753. If we decide to pursue this approach I'll update full-scan implementation with appropriate index selection. The main disadvantage is that it requires more changes compared with previous approaches.

So which way would you prefer to fix this? I'll prepare the appropriate PR after you decide.

brainopia:
@ibdknox This is the simplest fix for lookup with missing variables (albeit not performant). Plus I've changed default index for full-scan to `ave`.

I haven't pushed any other full-scan optimizations since in this implementation missing variables won't trigger full-scan. But if full-scan will be used in other contexts (asides from `****`) then please tell me and I'll be glad to push an adaptive index support to full-scan.

Closes #685.

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/715)

### DIFF
--- a/src/runtime/builder.ts
+++ b/src/runtime/builder.ts
@@ -299,6 +299,10 @@ function buildScans(block, context, scanLikes, outputScans) {
 
       if(!(entity || attribute || value || node)) {
         context.errors.push(errors.blankScan(block, scanLike));
+      } else {
+        entity = entity || context.createVariable();
+        attribute = attribute || context.createVariable();
+        value = value || context.createVariable();
       }
 
       let final = new join.Scan(scanLike.id + "|build", entity, attribute, value, node, scanLike.scopes);
--- a/src/runtime/join.ts
+++ b/src/runtime/join.ts
@@ -184,7 +184,7 @@ export class Scan {
     let solving = [];
     let solveNode = this.node !== undefined;
     let depth = solveNode ? 4 : 3;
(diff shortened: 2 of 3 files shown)


cmontella:
Ravil,

Sorry this sat around for a while, we'll review it shortly!

PR #722, opened 2017-01-06
Dot-notation for records

brainopia:
This PR adds support for code like `[#foo].tag` in a search section.

There is an issue that prevents it from working in an action section (scans should not generate proposals if a dependent variable is covered by some instance of `GenerateId` provider that hasn't been accepted yet), but I'll work on this in subsequent PR.

Let me know if there is anything you'd like me to change in the code.

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/722)

### DIFF
--- a/src/runtime/parser.ts
+++ b/src/runtime/parser.ts
@@ -567,14 +567,13 @@ export class Parser extends chev.Parser {
       // @TODO fill in from
       let from = [];
       let action = self.CONSUME(Action).image;
-      let actionKey = action;
       let scopes:any = ["session"];
       self.OPTION(() => { scopes = self.SUBRULE(self.scopeDeclaration) })
       self.activeScopes = scopes;
       self.currentAction = action;
       let statements = [];
       self.MANY(() => {
-        let statement = self.SUBRULE(self.actionStatement, [actionKey]) as any;
+        let statement = self.SUBRULE(self.actionStatement) as any;
         if(statement) {
           statements.push(statement);
           statement.scopes = scopes;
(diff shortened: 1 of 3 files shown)


ibdknox:
@joshuafcole @cmontella do either of you have any objections to this? I think I was the main opponent, but I'm not strongly opposed - more just a slight unease.

cmontella:
I think it'll make code a little harder to read in some cases, but it seems like a natural thing to want to do. I don't have any specific concerns though, so I think it's fine to allow this.

brainopia:
Oh, if you don't want this, I'll be fine without it :)

My reasoning was It seemed intuitive that if

```
foo = [#some-record]
foo.attribute
```

then

```
[#some-record].attribute
```

is expected to work too. But I admit that it may be far-fetched to apply this traditional reasoning to eve. 
Especially since my initial use-case was based on incorrect assumption that without this PR there is no way to simplify following code:

```
[#foo bar]
(not bar)
```

But I've learned about
```
[#foo not(bar)]
```
since then.

joshuafcole:
The big use case here is for inline queries in maths, e.g.

`[#person age < 18].height / [#teacher not(#professor)].height`

In principle it sounds like a good thing to support, but every time I write
an example of it I find it relatively hard to read. This is actually
something I was advocating for internally earlier, but after trying it out
with the logical syntax extension here, I'm not so sure. I can't think of a
particularly better syntax either. There's an argument to be made that we
should support it anyway for orthogonal completeness, but it doesn't seem
super compelling to me. I'd say I'm about a 2/10 against this. If you feel
more strongly that we should have it I'm okay with that.

brainopia:
Please correct me if I'm wrong in my following assumptions. It seems like the core team does not have strong arguments against this PR, but does have a minor gut feeling against it. And the team is ok to merge it if one person (who wrote the PR and is outside of the core team) strongly desires it. Is this correct?

In my opinion, seems like there are better ways to make the decision.
- A gut feeling of the whole core team should be enough to stop the PR.
- Community feedback could be gathered if the team is undecided/split. It'll hopefully provide an additional information that could be used by the team to make the final decision.
- If there is no proponent of PR on the team, then the burden of conviction lies on an outside person (that could include initiating a gathering of community feedback by that person). Until the team is convinced that a PR is' a good idea, it shouldn't be merged. A simple reason to close the PR from the core team like "not convinced it will help" is sufficient.

What do you think?

cmontella:
@brainopia I think it's a minor enough change that we should just allow it and see how it works. Like you said, it makes sens, and as josh said, it we might want it just as an option for completeness. I say let's merge it and see how it goes.

ibdknox:
@cmontella It should go the other way around. Anything we do that affects the actual language should be justified. That doesn't mean we won't eventually find a good reason for this, but we don't currently have one and its use *does* make things harder to read. As @brainopia the gut feeling is this probably shouldn't happen. I opened it up because @joshuafcole has previously expressed a desire for it, but he's lukewarm on it at best.

Let's close this for now. If someone comes up with a compelling justification for it, we'll revisit.

brainopia:
@ibdknox Sounds good to me, thanks for making a decision :+1:

brainopia:
@ibdknox by the way, I have another question.

Would you please check two first commits in the PR and tell me if you're interested in merging them. 

There are no changes in semantics/syntax, but in my opinion, it is a slight improvement of parser by removing a couple of redundant abstractions.

brainopia:
Scratch my question. It's irrelevant now :)

PR #725, opened 2017-01-07
enable source maps debugging

wilk:
It was impossible to use source maps debugging (with breakpoints on .ts sources) on Chrome (55) until I've explicitly set inlineSources to true

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "outDir": "build",
     "target": "es5",
     "sourceMap": true,
+    "inlineSources": true,
     "skipDefaultLibCheck": true,
     "forceConsistentCasingInFileNames": true,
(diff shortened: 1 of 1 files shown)


ibdknox:
@joshuafcole is this ok? It's certainly going to make the payload bigger, but maybe that's fine given our current dev-only state.

joshuafcole:
This will impact production packages too unless we disable it for
production builds. I think the better approach would be to include the
sourcemaps into the production package so it can be served separately still.

joshuafcole:
Hey @wilk, I'm going to close this PR since 0.2 is approaching EOL, If you're still interested, we can revisit the issue in 0.4!

wilk:
No problem @joshuafcole ;)

PR #735, opened 2017-01-12
add string find

ibdknox:
adds a `find[]` method that takes the following arguments:

- *text*: the base text to look in
- *subtext*: the text we're looking for
- *from* (optional): an index to start from
- *case-sensitive* (optional): whether or not the search should take case into account (default false)

And returns:

- *string-position*: position of the subtext in the string
- *result-index*: the index of this position in the results (e.g. 1 is the first occurrence of subtext)

### DIFF
--- a/src/runtime/providers/string.ts
+++ b/src/runtime/providers/string.ts
@@ -182,7 +182,6 @@ class Find extends Constraint {
     return text.indexOf(subtext, returns[0] - 1) === returns[0];
   }
 
-  // substring always returns cardinality 1
   getProposal(tripleIndex, proposed, prefix) {
     let proposal = this.proposalObject;
     let {args} = this.resolve(prefix);
--- a/src/runtime/providers/string.ts
+++ b/src/runtime/providers/string.ts
@@ -107,7 +107,6 @@ class Substring extends Constraint {
     let to = text.length;
     if (args[1] != undefined) from = args[1] - 1;
     if (args[2] != undefined) to = args[2];
-    console.log("test string", text.substring(from, to), from, to, returns[0]);
     return text.substring(from, to) === returns[0];
   }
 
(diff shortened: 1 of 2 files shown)


src/runtime/providers/string.ts
```diff
@@ -125,6 +124,90 @@ class Substring extends Constraint {
...
 
+class Find extends Constraint {
+  static AttributeMapping = {
+    "text": 0,
+    "subtext": 1,
+    "case-sensitive": 2,
+    "from": 3,
+  }
+  static ReturnMapping = {
+    "string-position": 0,
```
joshuafcole: This is going to be the codepoint index right now. I don't think that's a blocker for 6sq, but it probably needs to get an as property before we make it fully public.

src/runtime/providers/string.ts
```diff
@@ -125,6 +124,90 @@ class Substring extends Constraint {
...
+
+  returnType: "both" | "position";
+
+  constructor(id: string, args: any[], returns: any[]) {
+    super(id, args, returns);
+    if(this.returns[1] !== undefined && this.returns[0] !== undefined) {
+      this.returnType = "both"
+    } else if(this.returns[0] !== undefined) {
+      this.returnType = "position";
+    }
```
joshuafcole: I don't think it's very important, but it seems reasonable to want just the result-index if you're just counting occurrences or something.

src/runtime/providers/string.ts
```diff
@@ -125,6 +124,90 @@ class Substring extends Constraint {
...
+
+  test(prefix) {
+    let {args, returns} = this.resolve(prefix);
+    let text = args[Find.AttributeMapping["text"]];
+    let subtext = args[Find.AttributeMapping["subtext"]];
+    if(typeof text !== "string"|| typeof subtext !== "string") return false;
+    return text.indexOf(subtext, returns[0] - 1) === returns[0];
+  }
+
+  // substring always returns cardinality 1
```
joshuafcole: The code is right but this comment is copypasta.

test/strings.ts
```diff
@@ -169,10 +169,99 @@ test("length of multi-byte characters as code-points", (assert) => {
...
+      ["b", "at", "26"],
+      ["c", "at", "28"],
+      ["d", "at", "34"],
+    ],
+    remove: []
+  };
+  evaluate(assert, expected, `
+    ~~~
+    search
+      at = find[text: "I learned to play the Ukulele in Lebanon.", subtext: "le"]
```
joshuafcole: That sounds like a wonderful time and I hope to hear more about it.

src/runtime/providers/string.ts
ibdknox: you'd still end up needing an aggregate (either max or count), so I'm not sure that would buy you anything in that context

test/strings.ts
ibdknox: 🎸

src/runtime/providers/string.ts
ibdknox: removed

PR #738, opened 2017-01-13
Feature/multi doc

joshuafcole:
This is only intended for internal use at this stage.

It is a *very* rough PoC for multi-document support. It grabs *all* documents from the root workspace and evaluates them as a batch. The setup makes evaluation very expensive, but doesn't impact parsing time.

### DIFF
--- a/src/runtime/runtimeClient.ts
+++ b/src/runtime/runtimeClient.ts
@@ -168,7 +168,7 @@ export abstract class RuntimeClient {
         let documents = eveSource.fetchAll("root");
         for(let documentId in documents) {
           code += `# !!! ${documentId} !!! \n`;
-          code += (documentId === data.documentId ? this.lastParse.code : documents[documentId]) + "\n\n";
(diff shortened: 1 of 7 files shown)


src/runtime/runtimeClient.ts
```diff
@@ -160,6 +162,27 @@ export abstract class RuntimeClient {
       }
       this.send(JSON.stringify({type: "parse", generation: data.generation, text, spans, extraInfo, css: this.enabledCss()}));
     } else if(data.type === "eval") {
+      let parse = this.lastParse;
+      if(config.multiDoc) {
+        let code = "";
+        let documents = eveSource.fetchAll("root");
+        for(let documentId in documents) {
+          code += `# !!! ${documentId} !!! \n`;
+          code += (documentId === data.documentId ? this.lastParse.code : documents[documentId]) + "\n\n";
```
ibdknox: Why are you using `this.lastParse.code here`?

PR #741, opened 2017-01-19
Add simple scanning scanning support

joshuafcole:
We added a very simple prototype for performing naturally joined scans over a log of EAVNs (technically EAVNTFCs, but we completely ignore the TFC at present).

The code documentation should hopefully speak for itself in terms of implementation. If it doesn't please call that out in the CR.


src/runtime/runtime.ts
```diff
@@ -54,25 +54,313 @@ export class Interner {
...
+   * A field is unresolved if it is completely ignored by the scan or is an output of the scan.
+   */
+  fieldUnresolved(resolved:ResolvedEAVN, key: keyof ResolvedEAVN) {
+    return resolved[key] === IGNORE_REG || resolved[key] === undefined;
+  }
+
+  /**
+   * A field is not a static match if it is ignored, not a static field, or the input value does not match the static value.
+   */
+  notStaticMatch(input:Change, key: "e"|"a"|"v"|"n") {
```
brainopia: Seems like `"e"|"a"|"v"|"n"` type will come up quite often (`fieldUnresolved` as an example) and `keyof ResolvedEAVN` doesn't seem appropriate semantically here, should a new type be created for this?

brainopia: I'm far from a native speaker but isn't a name `notStaticMatch` a bit misleading and should be more like `staticNotMatch` or `staticValueNotMatch`? Forgive me for silly question :)

src/runtime/runtime.ts
```diff
@@ -54,25 +54,313 @@ export class Interner {
...
-console.log(GlobalInterner.intern("zomg"))
-console.log(GlobalInterner.intern("yo"))
+/**
+ * A change is a single changed EAV row of a changeset.
+ *  E.g., if we add [#person name: "josh"] then
+ *  (<1>, "tag", "person", ...) and
+ *  (<1>, "name", "josh", ...) are each separate changes.
+ */
+
+class Change {
```
brainopia: Could you document please what `count` means in the context of `Change`?

src/runtime/runtime.ts
brainopia: Disregard my first comment, I see that it's already done in the branch: https://github.com/witheve/Eve/pull/744/files#diff-677ffa07af969f1cb122adf871ee9fc8R88

joshuafcole: :+1: to the first comment.

joshuafcole: To the second, I agree that the name isn't great, but the alternatives are more disingenuous. What we're trying to capture is the situation where the field specified by `key` *is* a static value that doesn't match (hence, it's not a static match). Unfortunately `staticNotMatch` sounds very unnatural -- in English, it'd be preferred to choose the slightly more ambiguous name. `staticFieldNotMatch` is closer, but in every case you still need to already know what it does for the name to really make sense. I'm certainly willing to hear other alternatives, if you have any ideas!

src/runtime/runtime.ts
joshuafcole: We absolutely will once it does something -- right now we're prototyping how the incremental stuff will work, so things that are liable to change are more lightly documented than the more permanent bits. Just for your personal edification though (since you've been kind enough to read along!) `count` represents the multiplicity of the change--that is--how many of the change we're adding or removing.

`Change.fromValues("<1>", "tag", "person", 73, 1, 0 , 3)` would add three of the EAVN `("<1>", "tag", "person", 73)`, while `Change.fromValues("<1>", "tag", "person", 73, 1, 0 , -1)` would remove one of it.

These multiplicities are very important to the incremental model we're trying to adopt, which is based in McSherry's work with Differential Dataflow. It's super cool, and absolutely worth looking into if the subject interests you.

Issue #739, opened 2017-01-15
Unicode symbols as attributes.


Cormac-Williams:
When playing with issues around attributes I noticed that we may be running into a minor issue if the use of unicode symbols as attribute names continues.

This was with the gaussian function μ symbol, which apparently has multiple unicode symbols which do not compare directly.

Best way to solve would be either to change the attribute names to a textual equivalent
or to add all variants to the AttributeMapping file and handle with extra code in the constraint.

Regards,
Cormac.

cmontella:
I think the way to go here is keep it simple and change the attribute name to regular text. Using μ was a cool demonstration of unicode support in a function, but it's not very easy to type in the current editor. I think these attributes should be "mean" and "stdev" as they are commonly called.

Cormac-Williams:
Pull request in.

PR #742, opened 2017-01-19
Rename Gaussian Attributes to not use Unicode.

Cormac-Williams:
For https://github.com/witheve/Eve/issues/739

Once in I'll also post a change notice to the news-group.
Regards,
Cormac.


cmontella:
Thanks!

Issue #631, opened 2016-12-01
+ should not concat strings


cmontella:
For example

```
x = "foo" + "bar"
```

Then `x` should not be "foobar"

Cormac-Williams:
Do you wish to it replace by an explicit string concatenation operator, if so what? Or is string interpolation the only way to go?

joshuafcole:
Out of curiosity, what would differentiate the concatenation operator from
string interpolation? Internally, the concatenation operator is already
implemented, but we chose to only expose interpolation add its interface
because it seemed like a superset of functionality with a nice ux.

Cormac-Williams:
No particular difference. I was asking because I was considering putting together the pull to handle this.
There was discussion with @ibdknox on a recent pull re: this specific subject, my opinion as a programmer is that I like to have both, there's not much in it in terms of typing, it's more a new language making me learn something new. Since Eve is already so different that may not 

If you remove the current ability to add strings with the Math.ts Add function and don't replace with another operator then any

search
  x = "foo" + "bar"
commit @browser
[#div text:x]

block would filter, I'm fairly sure that would irritate some experienced programmers new to Eve ;-)

In fairness I had a play and have found a case of unexpected behaviour from the current concat version.

The following 

```
commit
  [#test type:"Num - Num" x:1 y:1]
  [#test type:"Str - Num" x:"1" y:1]
  [#test type:"Num - Str" x:1 y:"1"]
  [#test type:"Srt - Str" x:"1" y:"1"]
  [#test type:"Srt - Str" x:"0" y:"1"]
```

```
search
  tmp = [#test x y]
bind
  tmp.result := x + y
```

```
search
  tmp = [#test x y result]
bind @view
  [#value value:"{{tmp.type}} {{tmp.x}} + {{tmp.y}} = {{tmp.result}}"]
```

yields
Srt - Str 0 + 1 = 01
Num - Num 1 + 1 = 2
Str - Num 1 + 1 = 2
Num - Str 1 + 1 = 2
Srt - Str 1 + 1 = 2

Regards,
Cormac.

joshuafcole:
String interpolation and templating seem to be increasingly popular in
languages and frameworks used by webdevs these days, though I don't know
what the natural exposure for systems and native programmers is like here.
That said, I'm unsure that overloading `+` to do string concatenation as
well would be the right move. By making it similar but slightly different
we invite new users to trip over the little differences like lack of
implicit coercion. If, to avoid that, we chose to make it a separate
[...]

PR #743, opened 2017-01-19
Fix + should not concat strings

Cormac-Williams:
For https://github.com/witheve/Eve/issues/631 finally.

I've gone for the more conservative option of only filtering if one or both items cannot be converted to a number. 

Let me know if you want this code to filter and I'll change

search
a = "1"
b = "2"
x = a + b
commit @browser
[#div text:x]

Which will currently set x to 3.

### DIFF
--- a/test/math.ts
+++ b/test/math.ts
@@ -365,9 +365,9 @@ test("Test gaussian seed", (assert) => {
   evaluate(assert, expected, `
     ~~~
     search
-      g1 = gaussian[seed:0, σ:1.0, µ:0.0]
-      g2 = gaussian[seed:0, σ:1.0, µ:0.0]
-      g3 = gaussian[seed:1, σ:1.0, µ:0.0]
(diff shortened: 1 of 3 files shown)


cmontella:
Thanks for putting this together Cormac. On the topic of implicit conversion (treating a string as a number silently), I think that should be a failure as well.

test/math.ts
```diff
@@ -250,6 +249,95 @@ test("Test that string concatenation is still working after NaN change.", (asser
...
+      x =  a + b
+
+    bind
+      [result: x]
+    ~~~
+  `);
+  assert.end();
+});
+
+test("Test that plus with number + string does work.", (assert) => {
```
cmontella: This test should return nothing i.e. 1 + "1" is undefined

Cormac-Williams:
Done.

Cormac-Williams:
Done

Regards,
Cormac

cmontella:
Thanks Cormac!

PR #744, opened 2017-01-20
Generic join, function constraints, really bad indexes

ibdknox:
This lays out the interfaces for Constraints (which will be Scans, FunctionConstrains, Choose, Not, etc) and Indexes. Using the `Constraint` and `Index` interfaces we implemented generic join in JoinNode as well as a very simple ListIndex.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -1,3 +1,7 @@
+//------------------------------------------------------------------------
+// Runtime
+//------------------------------------------------------------------------
+
 type ID = number;
 type Multiplicity = number;
 
@@ -5,16 +9,24 @@ function createHash() {
   return Object.create(null);
 }
 
+function createArray() {
+  return [];
+}
+
 function isNumber(thing:any) {
   return typeof thing === "number";
 }
 
+//------------------------------------------------------------------------
+// Interning
+//------------------------------------------------------------------------
+
 export class Interner {
   strings: {[value:string]: ID|undefined} = createHash();
   numbers: {[value:number]: ID|undefined} = createHash();
-  IDs: (string|number)[] = [];
-  IDRefCount: number[] = [];
-  IDFreeList: number[] = [];
+  IDs: (string|number)[] = createArray();
+  IDRefCount: number[] = createArray();
+  IDFreeList: number[] = createArray();
   ix: number = 0;
 
   _getFreeID() {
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -69,6 +81,106 @@ export class Interner {
...
+        if(!seen[current]) {
+          seen[current] = true;
+          final.push([current]);
+        }
+      }
+    }
+
+    proposal.cardinality = final.length;
+    proposal.info = final;
+    proposal.forFields = forFields;
```
brainopia: If this proposal solves only for `forFields[0]`, why are all variable fields returned in the proposal?

src/runtime/runtime.ts
```diff
@@ -69,6 +81,106 @@ export class Interner {
...
+}
+
+//------------------------------------------------------------------------
+// Indexes
+//------------------------------------------------------------------------
+
+interface Index {
+  insert(change:Change):void;
+  propose(proposal:Proposal, e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction:number, round:number):Proposal;
+  resolveProposal(proposal:Proposal):any[][];
```
brainopia: Shouldn't it be `ID[][]`?

src/runtime/runtime.ts
```diff
@@ -1,20 +1,32 @@
...
+//------------------------------------------------------------------------
+
 type ID = number;
 type Multiplicity = number;
 
 function createHash() {
   return Object.create(null);
 }
 
+function createArray() {
```
brainopia: What's the idea behind `createArray`? Consistency with `createHash`, pretty name or something else entirely?

src/runtime/runtime.ts
```diff
@@ -288,86 +336,357 @@ class Scan {
...
    */
   applyInput(input:Change, prefix:ID[]) {
-    let resolved = this.resolve(prefix);
-
     // If this change isn't relevant to this scan, skip it.
-    if(this.notStaticMatch(input, "e")) return true;
-    if(this.notStaticMatch(input, "a")) return true;
-    if(this.notStaticMatch(input, "v")) return true;
-    if(this.notStaticMatch(input, "n")) return true;
+    if(this.notStaticMatch(input, "e")) return ApplyInputState.none;
```
brainopia: The function comment still documents returned values as true or false. Needs to be updated.

src/runtime/runtime.ts
```diff
@@ -288,86 +336,357 @@ class Scan {
...
+  pass,
+  fail,
+  none,
+}
+
+interface Constraint {
+  setup():void;
+  getRegisters():Register[];
+  applyInput(input:Change, prefix:ID[]):ApplyInputState;
+  propose(index:Index, prefix:ID[], transaction:number, round:number, results:any[]):Proposal;
```
brainopia: Is `results` argument necessary for `propose` function? It's not used anywhere yet, will it be?

src/runtime/runtime.ts
```diff
@@ -288,86 +336,357 @@ class Scan {
...
+  }
+
+  propose(index:Index, prefix:ID[], transaction:number, round:number, results:any[]):Proposal {
+    let {e,a,v,n} = this.resolve(prefix);
+    this.proposal.skip = false;
+    let proposal = index.propose(this.proposal, e, a, v, n, transaction, round);
+
+    let ix = 0;
+    for(let field of proposal.forFields) {
+      proposal.forRegisters[ix] = this[field] as Register;
```
brainopia: A number of `forFields` could vary between different runs (e.g. `listIndex.propose` currently returns all variable fields as `forFields` and that array will decrease each time a new variable is solved for). Therefore `forRegisters` could contain some left-overs from previous runs and should be cleaned up before re-used.

Do you want me to send a PR to change this?

src/runtime/runtime.ts
joshuafcole: We use these functions during development to make it easy to track allocations. This isn't particularly important outside of the evaluation loop, but inside it limiting allocations is critical to achieving consistent (let alone good!) performance.

src/runtime/runtime.ts
joshuafcole: Yes, this gets fixed in the future.

src/runtime/runtime.ts
joshuafcole: GenericJoin solves for a single register at a time, but a scan may be able to resolve multiple values at once (e.g., all the AV pairs for a given E). In this case, we must keep track of all of those related fields, since both the valid As and valid Vs are limited to the groups that are paired together.

src/runtime/runtime.ts
joshuafcole: You spotted a very obnoxious bug that we ended up fixing in the future. Good eye! Luckily this one broke things badly enough that it didn't go unnoticed for long. Unfortunately in the pursuit of fast code ugly and tough to write things like this have to be done sometimes. If you see any other issues like this, please don't hesitate to point them out -- a spare set of eyes really makes fixing these much easier!

src/runtime/runtime.ts
joshuafcole: We have some thoughts about using this for scenarios where computing the proposal is expensive. We could theoretically cache it in the results to be used if the proposal succeeds. Currently we use `proposal.info` for that, but it incurs a copying cost. However, the exact implementation of this is still up in the air and it doesn't really make sense to drag extra parameters that don't work yet around, so I agree that we should remove it until we need it. :+1:

PR #746, opened 2017-01-20
Transactions

ibdknox:
Add a transaction wrapper that fixpoints all our changes.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -550,7 +550,7 @@ interface Node {
    * See Scan.exec()
    * @NOTE: The result format is slightly different. Rather than a packed list of EAVNs, we instead return a set of valid prefixes.
    */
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -550,7 +550,7 @@ interface Node {
    * See Scan.exec()
    * @NOTE: The result format is slightly different. Rather than a packed list of EAVNs, we instead return a set of valid prefixes.
```
joshuafcole: This comment here is old

src/runtime/runtime.ts
```diff
@@ -698,6 +698,7 @@ class InsertNode implements Node {
 
     let change = new Change(resolved.e!, resolved.a!, resolved.v!, resolved.n!, transaction, round + 1, 1);
     console.log(""+change);
+    changes.push(change);
```
joshuafcole: We should probably just print the changes for the whole TX now.

joshuafcole:
Closed in favor of rebased #751

PR #747, opened 2017-01-20
Fix a required typescript version

brainopia:
`keyof` type is supported only since typescript 2.1
(but reliably compiles runtime.ts only since 2.1.4)

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/747)


joshuafcole:
Good catch! I'd been using a global tsc install.

PR #748, opened 2017-01-20
Refactor runtime functions

joshuafcole:
Mostly working*

This is still busted for interesting functions (where not all fields can be sourced from the `input` change) because (I think) our scans are being too greedy about using only values from the input rather than the index thanks to `acceptInput`. I added a useless `=` function as a PoC that seems okay for filtering, and + works (so long as it meets the requirement above).

```
search
  [age:age1]
  [age:age2]
  age1 > age2
commit
  [age1 age2]
```

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -2,6 +2,7 @@
 // Runtime
 //------------------------------------------------------------------------
 
+type RawValue = string|number;
 type ID = number;
 type Multiplicity = number;
 
@@ -13,7 +14,7 @@ function createArray() {
   return [];
 }
 
-function isNumber(thing:any) {
+function isNumber(thing:any): thing is number {
   return typeof thing === "number";
 }
 
@@ -24,7 +25,7 @@ function isNumber(thing:any) {
 export class Interner {
   strings: {[value:string]: ID|undefined} = createHash();
   numbers: {[value:number]: ID|undefined} = createHash();
-  IDs: (string|number)[] = createArray();
+  IDs: RawValue[] = createArray();
   IDRefCount: number[] = createArray();
   IDFreeList: number[] = createArray();
   ix: number = 0;
@@ -33,7 +34,7 @@ export class Interner {
     return this.IDFreeList.pop() || this.ix++;
   }
 
-  intern(value: (string|number)): ID {
+  intern(value: RawValue): ID {
     let coll;
     if(isNumber(value)) {
       coll = this.numbers;
@@ -52,7 +53,7 @@ export class Interner {
     return found;
   }
 
-  get(value: (string|number)): ID|undefined {
+  get(value: RawValue): ID|undefined {
     let coll;
     if(isNumber(value)) {
       coll = this.numbers;
@@ -62,7 +63,7 @@ export class Interner {
     return coll[value];
   }
 
-  reverse(id: ID): (string|number) {
+  reverse(id: ID): RawValue {
     return this.IDs[id];
   }
 
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -488,6 +488,7 @@ abstract class FunctionConstraint implements Constraint {
   apply: (... things: any[]) => undefined|(number|string)[]; // @FIXME: Not supporting multi-return yet.
   estimate?:(index:Index, prefix:ID[], transaction:number, round:number) => number
 
+  fieldNames:string[];
   proposal:Proposal = {cardinality:0, forFields: createArray(), forRegisters: createArray(), proposer: this};
   protected resolved:ResolvedFields = {};
   protected registers:Register[] = createArray();
@@ -495,7 +496,9 @@ abstract class FunctionConstraint implements Constraint {
   protected applyInputs:RawValue[] = createArray();
 
   setup() {
-    for(let fieldName of Object.keys(this.fields)) {
+    this.fieldNames = Object.keys(this.fields);
+
+    for(let fieldName of this.fieldNames) {
       let field = this.fields[fieldName];
       if(isRegister(field)) this.registers.push(field);
     }
@@ -514,7 +517,7 @@ abstract class FunctionConstraint implements Constraint {
   resolve(prefix:ID[]) {
     let resolved = this.resolved;
 
-    for(let fieldName of Object.keys(this.fields)) {
+    for(let fieldName of this.fieldNames) {
       let field = this.fields[fieldName];
       if(isRegister(field)) {
         resolved[fieldName] = prefix[field.offset];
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
+  constructor(public fields:ConstraintFieldMap) {
 
   }
 
+  name:string;
+  args:{[name:string]: string};
+  returns:{[name:string]: string};
+  argNames:string[];
+  returnNames:string[];
+  apply: (... things: any[]) => undefined|(number|string)[]; // @FIXME: Not supporting multi-return yet.
```
joshuafcole: Returns RawValue[]

joshuafcole: Add applyMany for the multi-return case.

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
+    for(let fieldName of Object.keys(this.fields)) {
+      let field = this.fields[fieldName];
+      if(isRegister(field)) this.registers.push(field);
+    }
+
+    for(let register of this.registers) {
+      this.registerLookup[register.offset] = true;
+    }
+
+    if(this.name === "+") console.log(this.name, this.registers);
```
joshuafcole: Remove this log

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
+      }
+    }
+
+    return resolved;
   }
 
-  args: any;
-  returns: any;
-  apply: (... things: any[]) => any;
+  isFilter():boolean {
```
joshuafcole: Dead code

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
-    // we should only attempt to propose if our args are filled and at least one
-    // return is not
-    this.proposal.skip = true;
-    return this.proposal;
+    let proposal = this.proposal;
+    proposal.forRegisters.length = 0;
+    let resolved = this.resolve(prefix);
+
+    // If none of our returns are unbound
+    // @NOTE: We don't need to worry about the filter case here, since he'll be
```
joshuafcole: You should probably finish when

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
+    // First we build the args array to provide the apply function.
+    let inputs = this.packInputs(resolved);
+
+    // Then we actually apply it and then unpack the outputs.
+    // @FIXME: We don't have any intelligent support for not computing unnecessary returns atm.
+    // @FIXME: We only support single-return atm.
+    let outputs = this.unpackOutputs(this.apply.apply(this, inputs));
+    if(!outputs) return results;
+
+    // Finally, if we had results, we create the result prefixes and pass them along.
```
joshuafcole: Make sure the bound return registers line up with our outputs too, rather than just ignoring them.

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
+      }
+      ix++;
+    }
+    results.push(result);
+
     return results;
   }
 
   accept(index:Index, prefix:ID[], transaction:number, round:number, solvingFor:Register[]):boolean {
+    let resolved = this.resolve(prefix);
```
joshuafcole: Actually use solvingFor

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
 
   acceptInput(index:Index, input:Change, prefix:ID[], transaction:number, round:number):boolean {
-    return true;
+    // @TODO: Implement the logic for sorting function constraints or re-accepting after scan constraints to ensure prefix is filled.
+    // @NOTE: Can we be smarter than solving for all registers here?
+    return this.accept(index, prefix, transaction, round, this.registers);
   }
+}
+
+class FunctionConstraintInstance extends FunctionConstraint {
```
joshuafcole: Just make FunctionConstraint not abstract since it doesn't buy us anything.

src/runtime/runtime.ts
```diff
@@ -459,84 +460,276 @@ interface Constraint {
...
 makeFunction({
-  name: "<",
+  name: ">",
   args: {a: "number", b: "number"},
-  returns: {c: "number"},
+  returns: {},
   apply: (a:number, b:number) => {
-    return a < b;
-  }});
+    return (a > b) ? [] : undefined;
```
joshuafcole: Make FILTER_PASS / FILTER_FAIL constant for filters

src/runtime/runtime.ts
```diff
@@ -573,13 +767,31 @@ class JoinNode implements Node {
...
+    let ix = inputState.constraintIx;
+    // if we're not starting at the first constraint, that means we've applied
+    // one already, we need to unapply it in order to make sure things work out
+    if(ix !== 0) {
+      let prevConstraint = constraints[ix - 1];
+      for(let register of prevConstraint.getRegisters()) {
+        prefix[register.offset] = undefined;
+      }
+    }
+    for(let len = constraints.length; ix < len; ix++) {
```
joshuafcole: This logic is insufficient -- we need the full combinations of valid inputs.

src/runtime/runtime.ts
```diff
@@ -647,33 +861,44 @@ class JoinNode implements Node {
...
+    this.inputState.constraintIx = 0;
+    // we need to apply input for each constraint individually and then run genericJoin to
+    // see if it produces any output. This handles the case where a single input might apply
+    // to several constraints in the block and we need to make sure we cover every permutation.
+    // @TODO: handle the case where all the constraints have applied a change.
+    this.applyInput(input, prefix);
+    let ok = this.inputState.state;
+    let ix = this.inputState.constraintIx;
+    let foo = 0;
+    while(ix < this.constraints.length) {
```
joshuafcole: This needs to be re-re-worked to happen over the set of valid input combinations.

src/runtime/runtime.ts
```diff
@@ -459,84 +460,273 @@ interface Constraint {
...
-  // @TODO
   getRegisters() {
-    return createArray();
+    return this.registers;
+  }
+
+  resolve(prefix:ID[]) {
+    let resolved = this.resolved;
+
+    for(let fieldName of Object.keys(this.fields)) {
```
cmontella: Object.keys is expensive. We should precompute the fields.

src/runtime/runtime.ts
```diff
@@ -647,33 +855,44 @@ class JoinNode implements Node {
...
+    let didSomething = false;
+    this.inputState.constraintIx = 0;
+    // we need to apply input for each constraint individually and then run genericJoin to
+    // see if it produces any output. This handles the case where a single input might apply
+    // to several constraints in the block and we need to make sure we cover every permutation.
+    // @TODO: handle the case where all the constraints have applied a change.
+    this.applyInput(input, prefix);
+    let ok = this.inputState.state;
+    let ix = this.inputState.constraintIx;
+    let foo = 0;
```
cmontella: remove foo

PR #749, opened 2017-01-20
Fix Interner.release for the first interned object

brainopia:
A first interned object will have an id of 0. Therefore `if(!id) return` will wrongfully terminate release of it.
Also slightly faster :D

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/749)


joshuafcole:
Thanks for taking such a close look @brainopia! :)

PR #750, opened 2017-01-20
Fix a leak in Interner.release

brainopia:
Currently, each time an interned number is released an extra slot is allocated in interned strings storage (with value of undefined) which won't be reclaimable and vice-versa.

Either a check on an interned value type should be added or `delete` could be used (to prevent branching while fixing the issue).

---
This change is [Reviewable](https://reviewable.io/reviews/witheve/eve/750)

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -73,8 +73,13 @@ export class Interner {
     this.IDRefCount[id]--;
     if(!this.IDRefCount[id]) {
       let value = this.IDs[id];
-      this.numbers[value as number] = undefined;
-      this.strings[value as string] = undefined;
+      let coll;
+      if(isNumber(value)) {
+        coll = this.numbers;
+      } else {
+        coll = this.strings;
+      }
+      coll[value] = undefined;
       this.IDFreeList.push(id);
     }
   }


joshuafcole:
This is a problem, but we're hesitant to use the `delete` keyword -- a common pattern in performance degradations we've seen is that the `delete` key can end up de-opting a large segment of the stack of functions containing it. Let's just branch on the value type when we remove.

src/runtime/runtime.ts
```diff
@@ -60,8 +60,8 @@ export class Interner {
     this.IDRefCount[id]--;
     if(!this.IDRefCount[id]) {
       let value = this.IDs[id];
-      this.numbers[value as number] = undefined;
-      this.strings[value as string] = undefined;
+      delete this.numbers[value as number];
```
joshuafcole: Let's branch instead.

brainopia:
@joshuafcole Thank you for feedback. I've fixed the `delete` issue.

By the way, I wonder why can't numbers and strings be stored together? It'll simplify all these conditionals and we already depend on the type of argument for `Interner` to work correctly.

joshuafcole:
If we did then 1 and "1" would be the same value, since JS coerces
numerical object keys to strings. Will take a look when I get to a computer!

PR #752, opened 2017-01-21
Add variadic function constraint support and gen-id

joshuafcole:


src/runtime/runtime.ts
```diff
@@ -735,6 +790,30 @@ makeFunction({
...
+  name: "eve-internal/gen-id",
+  args: {},
+  variadic: true,
+  returns: {result: "string"},
+  apply: (values:RawValue[]) => {
+    // @FIXME: This is going to be busted in subtle cases.
+    // If a record exists with a "1" and 1 value for the same attribute, they'll collapse for gen-id, but won't join elsewhere.
+    // This means aggregate cardinality will disagree with action node cardinality.
+
+    return [values.join("|")];
```
cmontella: What happens when the value is a pipe?

joshuafcole: I think we should still be fine because node id will be included, and nodes can only output a fixed pattern. It is a little hokey though. At least requires a comment.

PR #754, opened 2017-01-22
performance tweaks, add a hash index

ibdknox:
This pulls indexes out into their own file so we can have a few different types. It also fixes a couple of show-stopping performance issues, reduces allocations a fair amount, and generally gets us into pretty good performance territory.

Before these changes the test case at the bottom took 16s, it's now ~300ms. The previous runtime couldn't do more than 3k updates (this does 10k) and hit ~75s doing so.

### DIFF
--- a/src/runtime/indexes.ts
+++ b/src/runtime/indexes.ts
@@ -9,6 +9,9 @@ function isResolved(field:ResolvedValue): field is ID {
   return field !== undefined && field !== IGNORE_REG;
 }
 
+// This function sums the counts of a packed array of node,
+// transcation, round, count (ntrc) up to the given transaction and
+// round based on the partial order (t1 <= t2 && r1 <= r2)
 function sumTimes(ntrcArray:number[], transaction:number, round:number) {
   if(!ntrcArray) return 0;
   let total = 0;
@@ -42,7 +45,7 @@ export class ListIndex implements Index {
     return proposal.info;
   }
 
-  propose(proposal:Proposal, e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction = Infinity, round = Infinity) {
+  propose(proposal:Proposal, e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction:number, round:number) {
     let final = createArray("indexProposeResults") as ID[][];
     let forFields:EAVNField[] = createArray("indexForFields");
     let seen = createHash();
@@ -73,7 +76,7 @@ export class ListIndex implements Index {
     return proposal;
   }
 
-  check(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction = Infinity, round = Infinity):boolean {
+  check(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction:number, round:number):boolean {
     for(let change of this.changes) {
       if((e === undefined || e === IGNORE_REG || e === change.e) &&
          (a === undefined || a === IGNORE_REG || a === change.a) &&
@@ -87,7 +90,7 @@ export class ListIndex implements Index {
     return false;
   }
 
-  get(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction = Infinity, round = Infinity):EAVN[] {
+  get(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction:number, round:number):EAVN[] {
     let final = createArray() as EAVN[];
     for(let change of this.changes) {
       if((e === undefined || e === IGNORE_REG || e === change.e) &&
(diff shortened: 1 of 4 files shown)


src/runtime/indexes.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    } else {
+      forFields[0] = fieldB;
+      proposal.info = Object.keys(bIx);
+      proposal.cardinality = proposal.info.length;
+      return proposal;
+    }
+  }
+
+  walkCheck(index:any, a:ResolvedValue, b:ResolvedValue, c:ResolvedValue, n:ResolvedValue, transaction:number, round:number):boolean {
```
joshuafcole: This guy deserves a comment along the lines of "check that the deepest resolved level of the index contains at least one entry." Took me a minute to work it out.

src/runtime/indexes.ts
```diff
@@ -0,0 +1,272 @@
...
+  check(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction = Infinity, round = Infinity):boolean {
+    if(isResolved(e)) {
+      return this.walkCheck(this.eavIndex, e, a, v, n, transaction, round);
+    } else if(isResolved(a)) {
+      return this.walkCheck(this.aveIndex, a, v, e, n, transaction, round);
+    }
+    return true;
+  }
+
+  get(e:ResolvedValue, a:ResolvedValue, v:ResolvedValue, n:ResolvedValue, transaction = Infinity, round = Infinity):EAVN[] {
```
joshuafcole: I assume this isn't implemented yet? If you stick `throw new Error("Implement me, please!")` in there it won't complain about not having a valid return value.

joshuafcole: Same deal if it's intended to be virtual, changing the message accordingly.

src/runtime/indexes.ts
```diff
@@ -0,0 +1,272 @@
...
+    let final = createArray() as EAVN[];
+    return final;
+  }
+}
+
+
+export class MatrixIndex implements Index {
```
joshuafcole: Should stick a comment above this guy marking him as not being usable yet. It's pretty obvious when in this file, but from other files there'll be no indication without a doc comment. Otherwise, you can drop the export til its ready for showtime.

src/runtime/runtime.ts
```diff
@@ -894,7 +818,7 @@ class JoinNode implements Node {
     return affectedConstraints;
   }
 
-  applyCombination(input:Change, prefix:ID[], transaction:number, round:number, results:ID[][]) {
+  applyCombination(index:Index, input:Change, prefix:ID[], transaction:number, round:number, results:ID[][]) {
```
joshuafcole: Does this guy actually use the index anywhere?

src/runtime/runtime.ts
```diff
@@ -961,7 +886,7 @@ class JoinNode implements Node {
     let {proposer} = bestProposal;
     // We have to slice here because we need to keep a reference to this even if later
```
joshuafcole: s/slice/copy

src/runtime/runtime.ts
```diff
@@ -1177,47 +1104,67 @@ let blocks:Block[] = [
...
+  //   new InsertNode(GlobalInterner.intern("adds-to"), GlobalInterner.intern("age1"), age1Reg, GlobalInterner.intern(76)),
+  //   new InsertNode(GlobalInterner.intern("adds-to"), GlobalInterner.intern("age2"), age2Reg, GlobalInterner.intern(76)),
+  //   new InsertNode(GlobalInterner.intern("adds-to"), GlobalInterner.intern("result"), resultReg, GlobalInterner.intern(76)),
+  // ]),
 
 ];
 
-let index = new ListIndex();
-let transaction = 0;
+export function doIt() {
```
joshuafcole: Palpatine Approves

src/runtime/runtime.ts
ibdknox: it passes it on to both presolveCheck and genericJoin. The only reason this worked before was because we had a global named "index" :(

src/runtime/indexes.ts
```diff
@@ -0,0 +1,272 @@
...
+
+//------------------------------------------------------------------------
+// Utils
+//------------------------------------------------------------------------
+
+function isResolved(field:ResolvedValue): field is ID {
+  return field !== undefined && field !== IGNORE_REG;
+}
+
+function sumTimes(ntrcArray:number[], transaction:number, round:number) {
```
ibdknox: node, transaction, round, count

src/runtime/runtime.ts
ibdknox: It is done.

src/runtime/runtime.ts
joshuafcole: whups, we need to be more disciplined and name globals in a way that they won't get picked up.

joshuafcole:
:ship: :shallow_pan_of_food:

PR #756, opened 2017-01-24
Refactor reorganize sections

joshuafcole:
General cleanup and documentation pass to make following along with the refactor easier for the community.

This is just a first pass, but I wanted to make sure there was enough structure and documentation to the code for the intrepid to get the gist of what's going on.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -789,6 +789,31 @@ interface Node {
   exec(index:Index, input:Change, prefix:ID[], transaction:number, round:number, results?:ID[][], changes?:Change[]):boolean;
 }
 
+/**
+ * The JoinNode implements generic join across multiple constraints.
+ * Since our system is incremental, we need to do something slightly fancier than we did in the previous runtime.
+ * For each new change that enters the system, we ask each of our constraints whether they are capable of producing
+ * a new result. In the case where a single constraint can, we presolve that constraint and then run the rest normally,
+ * limited to only producing results that match the first constraint. However, if multiple constraints might apply the input,
+ * we need to run for each *combination* of heads. E.g.:
+ *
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -784,6 +789,31 @@ interface Node {
...
 
+/**
+ * The JoinNode implements generic join across multiple constraints.
+ * Since our system is incremental, we need to do something slightly fancier than we did in the previous runtime.
+ * For each new change that enters the system, we ask each of our constraints whether they are capable of producing
+ * a new result. In the case where a single constraint can, we presolve that constraint and then run the rest normally,
+ * limited to only producing results that match the first constraint. However, if multiple constraints might apply the input,
+ * we need to run for each *combination* of heads. E.g.:
+ *
+ * Given a join node with constraints [A, B, C, and D], where A and C can both apply the input, we must combine the results of
```
ibdknox: I think you meant  A and D apply

src/runtime/runtime.ts
```diff
@@ -784,6 +789,31 @@ interface Node {
...
+ * the following computations to get the full result set:
+ * Apply A -> Do {B, C, D}
+ * Apply {A, D} -> Do {B, C}
+ * Apply {D} -> Do {A, B, C}
+ *
+ * We calculate this using the power set in exec.
+ *
+ * We then apply each of these combinations by running a genericJoin over the remaining unresolved registers.
```
ibdknox: the comment lines here are all pretty long, probably should flow them onto another line so no scrolling is necessary to read.

joshuafcole: I usually shoot for 150char columns, I know some people still stick to 80 but that seems very short to me. Where do you want us to be?

PR #757, opened 2017-01-25
JS DSL

ibdknox:
This adds the start of a JS DSL for the new runtime, allowing you to write Eve programs like so:

```typescript
let foo = new Program("foo");

foo.block("cool story", (find:any, record:any, lib:any) => {
  let person = find("person");
  let text = `name: ${person.name}`;
  return [
    record("html/div", {person, text})
  ]
})

foo.input(runtime.createChangeSet([1, "name", "chris", 1], [1, "tag", "person", 1]))
```

### DIFF
--- a/src/runtime/dsl.ts
+++ b/src/runtime/dsl.ts
@@ -1,0 +1,205 @@
+declare var Proxy:new (obj:any, proxy:any) => any;
+declare var Symbol:any;
+
+function concat(...strs:string[]) {
+  return {};
+}
+
+class FunctionCall {
+  constructor(public path:string[], public args:any[]) {}
+}
+
+class Variable {
+  constructor(public name:string) { }
+}
+
+class EveRecord {
+  constructor(public groups:string[], public fields:any) {
+    fields.groups = groups;
+  }
+
+  proxy() {
+    return new Proxy(this, {
+      get: function(obj:any, prop:string) {
+        let found = obj[prop] || obj.fields[prop];
+        if(prop === Symbol.toPrimitive) return () => {
+          return "uh oh";
+        }
+        if(!found) {
+          found = obj.fields[prop] = new Variable(prop);
+        }
+        return found;
+      }
+    })
+  }
+}
+
+class Program {
+  constructor(public name:string) {}
+
+  transformBlockCode(code:string):string {
+
+    let hasChanged = true;
+    let infixParam = "((?:(?:[a-z0-9_\.]+(?:\\[\".*?\"\\])?)+(?:\\(.*\\))?)|\\(.*\\))";
+    let stringPlaceholder = "(____[0-9]+____)";
+
+    let strings:string[] = [];
+    code = code.replace(/"(?:[^"\\]|\\.)*"/gi, function(str) {
+      strings.push(str);
+      return "____" + (strings.length - 1) + "____";
+    })
+
+    let stringAddition = new RegExp(`(?:${infixParam}\\s*\\+\\s*${stringPlaceholder})|(?:${stringPlaceholder}\\s*\\+\\s*${infixParam})`,"gi");
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(stringAddition, (str, left, right, left2, right2) => {
+        hasChanged = true;
+        if(left === undefined) {
+          left = left2;
+          right = right2;
+        }
+        left = this.transformBlockCode(left);
+        right = this.transformBlockCode(right);
+        strings.push(`fn.string.concat(${left}, ${right})`);
+        return "____" + (strings.length - 1) + "____";
+      })
+    }
+
+    // foo(..) -> fn.foo(..)
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(/([a-z0-9\._]+\(.*\))/gi, (str, fun) => {
+        if(!fun.match(/fn\.|^record\(/)) {
+          hasChanged = true;
+          return "fn." + fun;
+        }
+        return fun;
+      });
+    }
+
+    let multiply = new RegExp(`${infixParam}\\s*(\\*|\\/)\\s*${infixParam}`, "gi");
+    // a * b -> fn.math["*"](a, b)
+    // a / b -> fn.math["/"](a, b)
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(multiply, (str, left, op, right) => {
+        hasChanged = true;
+        left = this.transformBlockCode(left);
+        right = this.transformBlockCode(right);
+        strings.push(`fn.math["${op}"](${left}, ${right})`)
+        return "____" + (strings.length - 1) + "____";
+      });
+    }
+    // a + b -> fn.math["+"](a, b)
+    // a - b -> fn.math["-"](a, b)
+    let add = new RegExp(`${infixParam}\\s*(\\+|\\-)\\s*${infixParam}`, "gi");
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(add, (str, left, op, right) => {
+        hasChanged = true;
+        left = this.transformBlockCode(left);
+        right = this.transformBlockCode(right);
+        strings.push(`fn.math["${op}"](${left}, ${right})`)
+        return "____" + (strings.length - 1) + "____";
+      });
+    }
+    // a > b -> fn.compare[">"](a, b)
+    let compare = new RegExp(`${infixParam}\\s*(>|>=|<|<=|!=|==)\\s*${infixParam}`, "gi");
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(compare, (str, left, op, right) => {
+        hasChanged = true;
+        left = this.transformBlockCode(left);
+        right = this.transformBlockCode(right);
+        strings.push(`fn.compare["${op}"](${left}, ${right})`)
+        return "____" + (strings.length - 1) + "____";
+      });
+    }
+
+    let lastReturn = code.lastIndexOf("return");
+    code = code.replace(/record\(/gi, (str, pos) => {
+      if(pos > lastReturn) {
+        return "output(";
+      }
+      return str;
+    });
+
+    hasChanged = true;
+    while(hasChanged) {
+      hasChanged = false;
+      code = code.replace(/____([0-9]+)____/gi, function(str, index:string) {
+        let found = strings[parseInt(index)];
+        if(found) hasChanged = true;
+        return found || str;
+      })
+    }
+    return code;
+  }
+
+  block(name:string, func:any) {
+    let info = {records: [] as EveRecord[], variables: [], functions: [] as FunctionCall[]};
+    let recordFunc = (...args:any[]) => {
+      let lastArg = args[args.length - 1];
+      let proxied:any = {};
+      let groups;
+      if(typeof lastArg === "object") {
+        proxied = lastArg;
+        groups = args.slice(0, args.length - 1);
+      } else {
+        groups = args.slice(0, args.length);
+      }
+      // console.log("GROUPS", groups);
+      let rec = new EveRecord(groups, proxied);
+      info.records.push(rec);
+      return rec.proxy();
+    }
+    let output = (...args:any[]) => {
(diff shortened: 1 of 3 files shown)


src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+//--------------------------------------------------------------------
+
+// There don't seem to be TypeScript definitions for these by default,
+// so here we are.
+declare var Proxy:new (obj:any, proxy:any) => any;
+declare var Symbol:any;
+
+import {RawValue, Register, isRegister, GlobalInterner, Scan, IGNORE_REG, ID,
+        InsertNode, Node, Constraint, FunctionConstraint} from "./runtime";
+import * as runtime from "./runtime";
```
joshuafcole: Out of curiosity, why directly import everything but a couple of the classes?

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+
+//--------------------------------------------------------------------
+// DSLVariable
+//--------------------------------------------------------------------
+
+type DSLValue = RawValue|Register;
+class DSLVariable {
+  static CURRENT_ID = 0;
+  id: number;
+  constructor(public name:string, public value:DSLValue = new Register(-1)) {
```
joshuafcole: I assume the case with no value provided is an error? We should be more explicit about this if so.

joshuafcole: Seems like it's used intentionally that way elsewhere, which means it's not an error case but a weird sentinel value? We should consider expanding the type definition rather than having a malformed register.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+
+//--------------------------------------------------------------------
+// DSLRecord
+//--------------------------------------------------------------------
+
+class DSLRecord {
+  record: DSLVariable;
+  output: boolean;
+  fields: any;
+  constructor(public block:DSLBlock, public tag:string[], initial:any) {
```
joshuafcole: Initial should have a comment describing it, or be more descriptively named (attributes, perhaps?).

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+          found = new DSLVariable(prop);
+          obj.fields[prop] = [found];
+          this.block.registerVariable(found);
+        } else {
+          found = found[0];
+        }
+        return found;
+      },
+      set: (obj:any, prop:string, value:any) => {
+        if(prop === "output") {
```
joshuafcole: What's the significance of output here?

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+          return true;
+        }
+        if(!obj.fields[prop]) {
+          if(value.constructor !== Array) {
+            value = [value];
+          }
+          obj.fields[prop] = value;
+          return true;
+        }
+        this.block.equivalence(obj.fields[prop][0], value);
```
joshuafcole: What does this array contain? Is it a tuple tracking some metadata rather than a list?

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+    code = code.replace(/function\s*\((.*)\)\s*\{/, function(str:string, args:string) {
+      functionArgs.push.apply(functionArgs, args.split(",").map((str) => str.trim()));
+      return "";
+    });
+    // trim the final } since we removed the function bit
+    code = code.substring(0, code.length - 1);
+    code = this.transformBlockCode(code, functionArgs);
+    let neueFunc = new Function(functionArgs[0], functionArgs[1], functionArgs[2], code);
+    neueFunc(this.find, this.record, this.generateLib());
+    this.prepare();
```
joshuafcole: Prepare

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  generateLib() {
```
joshuafcole: This is cute. Where do we handle the situation where you try to invoke a non-existent expression though? Do we want to leave that til runtime in the off chance that it'll be added? One of the annoyances of the piggy-backed layer is that we don't really have  consistent access to development time feedback anymore, I guess, so erroring earlier doesn't buy us much. :/

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+        apply: (target:any, targetThis:any, args:any[]) => {
+          let func = new DSLFunction(this, path, args);
+          this.functions.push(func);
+          return func;
+        }});
+    }
+    return new Proxy({}, {get:fnGet});
+  }
+
+  find = (...args:any[]) => {
```
joshuafcole: Worth sticking a comment here along the lines of. `Find takes a list of tags to search for, optionally followed by a map of attributes.` or some such.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+    }
+    // console.log("tag", tag);
+    let rec = new DSLRecord(this, tag, proxied);
+    this.records.push(rec);
+    return rec.proxy();
+  }
+
+  record = (...args:any[]) => {
+    let out = this.find.apply(null, args);
+    out.output = true;
```
joshuafcole: Oh, I get the output thing now. The same guy is used both for search and commit. We should add some documentation somewhere nonetheless.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+      return a.value;
+    } if(a instanceof DSLRecord) {
+      return a.record.value;
+    } if(a instanceof DSLFunction) {
+      return a.returnValue.value;
+    }
+    return a;
+  }
+
+  equivalence(a:any, b:any) {
```
joshuafcole: Let's spare a couple comments talking about the cases here.

Each value is either:
 - a static (interned) eve value
 - a record
 - a register (variable)

In the first case direct comparison is sufficient to test equality. In the second, etc. etc.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+    } else if(aValue !== bValue) {
+      throw new Error(`Trying to equivalence two static values that aren't the same: ${aValue} and ${bValue}`);
+    }
+  }
+
+  prepare() {
+    let functions = this.functions.slice() as (DSLFunction | undefined)[];
+    // look through the functions for equivalences before we do anything else.
+    let ix = 0;
+    for(let func of functions) {
```
joshuafcole: Need some clarification on this loop.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+      if(!func || func.path[1] !== "==") continue;
+      this.equivalence(func.args[0], func.args[1]);
+      functions[ix] = undefined;
+      ix++;
+    }
+    let registerIx = 0;
+    for(let id in this.variableLookup) {
```
joshuafcole: I see, so it really will be a register, we just haven't assigned it yet. Let's maybe add a constant for UNASSIGNED_REG or something?

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+      if(!compiled) continue;
+      for(let item of compiled) {
+        if(item instanceof Scan || item instanceof FunctionConstraint) {
+          constraints.push(item);
+        } else {
+          nodes.push(item as Node);
+        }
+      }
+    }
+    nodes.unshift(new runtime.JoinNode(constraints))
```
joshuafcole: I know we don't have aggregates yet, but I assume once we do we'll need to be a little smarter here and stratify across those boundaries into separate JoinNodes?

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+          right = right2;
+        }
+        left = this.transformBlockCode(left, functionArgs);
+        right = this.transformBlockCode(right, functionArgs);
+        strings.push(`${libArg}.eve.internal.concat(${left}, ${right})`);
+        return "____" + (strings.length - 1) + "____";
+      })
+    }
+
+    let multiply = new RegExp(`${infixParam}\\s*(\\*|\\/)\\s*${infixParam}`, "gi");
```
joshuafcole: Since this isn't in a hot path, can we maybe make a helper function to cut the repetition down a bit? The extra noise makes it a bit hard for me to wade through.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+      hasChanged = false;
+      code = code.replace(add, (str, left, op, right) => {
+        hasChanged = true;
+        left = this.transformBlockCode(left, functionArgs);
+        right = this.transformBlockCode(right, functionArgs);
+        strings.push(`${libArg}.math["${op}"](${left}, ${right})`)
+        return "____" + (strings.length - 1) + "____";
+      });
+    }
+    // a > b -> fn.compare[">"](a, b)
```
joshuafcole: We'll need to support != and <= and friends still.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+    console.log(trans.changes.map((change, ix) => `    <- ${change}`).join("\n"));
+  }
+}
+
+//--------------------------------------------------------------------
+// Testing
+//--------------------------------------------------------------------
+
+let foo = new Program("foo");
+
```
joshuafcole: Would like to see a few more test cases as well, both to show the ergonomics of the DSL and as part of our smoke testing until we get real tests rewired.

src/runtime/dsl.ts
```diff
@@ -0,0 +1,431 @@
...
+    let infixParam = "((?:(?:[a-z0-9_\.]+(?:\\[\".*?\"\\])?)+(?:\\(.*\\))?)|\\(.*\\))";
+    let stringPlaceholder = "(____[0-9]+____)";
+
+    let strings:string[] = [];
+    code = code.replace(/"(?:[^"\\]|\\.)*"/gi, function(str) {
+      strings.push(str);
+      return "____" + (strings.length - 1) + "____";
+    })
+
+    let stringAddition = new RegExp(`(?:${infixParam}\\s*\\+\\s*${stringPlaceholder})|(?:${stringPlaceholder}\\s*\\+\\s*${infixParam})`,"gi");
```
joshuafcole: Let's show representative examples of the patterns we expect to catch here, like we do for the infixes below.

src/runtime/runtime.ts
```diff
@@ -960,27 +960,46 @@ class JoinNode implements Node {
...
-      let ix = 0;
-      for(let register of forRegisters) {
-        prefix[register.offset] = result[ix];
-        ix++;
-      }
-      for(let constraint of constraints) {
-        if(constraint === proposer) continue;
-        if(!constraint.accept(index, prefix, transaction, round, forRegisters)) {
-          continue resultLoop;
+    let resolved:any[] = proposer.resolveProposal(index, prefix, bestProposal, transaction, round, proposedResults);
```
joshuafcole: What's up with this guy?

src/runtime/dsl.ts
ibdknox: I started out directly importing everything and then realized I was going to only use several of these once, so it was a bit of a waste. Probably worth cleaning up some more of these that are only used once just for clarity.

src/runtime/dsl.ts
ibdknox: We find out about this when we try to compile the function, which as far as the user is considered is the same time as the block is declared.

src/runtime/dsl.ts
joshuafcole: Makes sense.

src/runtime/dsl.ts
ibdknox: correct

src/runtime/dsl.ts
ibdknox: it does, that was just an example

src/runtime/dsl.ts
ibdknox: I've spent a good portion of today getting testing back. I wanted to get this guy in so that we can just have real test instead of doing the stuff at the end of files like we are now.

src/runtime/runtime.ts
ibdknox: The resolution can be either an `ID[]` or an `ID[][]`, but the typechecker didn't feel like cooperating based on the if statement on the next line

joshuafcole: We should consider giving this a distinct name since we use `resolved` to mean something else elsewhere.

PR #758, opened 2017-01-25
Arena allocate function interns

ibdknox:
Get the interner to arena allocate ids for functions so that we don't unnecessarily accumulate them. This also sets up to be able to release IDs from removes if we think that's the right thing to do. If we do end up doing that though, we have to be aware that holding onto a change after a transaction has finished will lead to potentially unresolvable values (or worse, incorrectly resolved values).

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -83,9 +83,18 @@ export class Interner {
   IDRefCount: number[] = createArray();
   IDFreeList: number[] = createArray();
   ix: number = 0;
+  arenas: {[arena:string]: ID[]} = createHash();
+
+  constructor() {
+    this.arenas["functionOutput"] = createArray("interner-arena-array");
+  }
 
   _getFreeID() {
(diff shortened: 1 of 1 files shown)


src/runtime/runtime.ts
```diff
@@ -1140,8 +1190,12 @@ export class Transaction {
       index.insert(change);
       changeIx++;
     }
-  }
 
+    // Once the transaction is effectively done, we need to clean up after ourselves. We
+    // arena allocated a bunch of IDs related to function call outputs, which we can now
+    // safely release.
+    GlobalInterner.releaseArena("functionOutput");
```
joshuafcole: Using a single fixed arena is fine until we go concurrent. We should consider just using some unique info off the transaction to create our arena.

ibdknox: the only problem is that we've have to pass that information down through everything somehow that isn't insane. Since we're pretty far off from having multiple threads (and maybe never will?). I think we can worry about that later.

PR #759, opened 2017-01-26
Testing in refactor branch

ibdknox:
Now that we have a reasonable way to write Eve using the DSL, it's time for real testing infrastructure!


src/runtime/runtime.ts
```diff
@@ -402,20 +412,30 @@ export class FunctionConstraint implements Constraint {
...
   }
 
+  static filter = false;
   static variadic = false;
+  static fetchInfo(name:string):typeof FunctionConstraint {
+    return FunctionConstraint.registered[name];
+  }
 
-  static create(name:string, fields:ConstraintFieldMap, restFields:(ID|Register)[] = createArray()):FunctionConstraint|undefined {
+  static create(name:string, fields:ConstraintFieldMap, restFields:(ID|Register)[] = createArray(), packRestIntoFields?:boolean):FunctionConstraint|undefined {
```
joshuafcole: Why do we need packRestIntoFields?

test/util.ts
```diff
@@ -0,0 +1,119 @@
+import {Program} from "../src/runtime/dsl";
+import * as Runtime from "../src/runtime/runtime";
+import * as test from "tape";
+
+// You can specify changes as either [e,a,v] or [e,a,v,round,count];
+export type EAVArray = [Runtime.RawValue, Runtime.RawValue, Runtime.RawValue];
```
joshuafcole: Personal preference, but I think it's clearer to refer to arrays with a fixed set of fields as tuples, since it's a lot more obvious that `myTuple.push(...)` is funky compared to `myArray.push(...)`

test/util.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  let all = ins.concat(outs);
+  let {changes} = program.input(ins);
+  let msg = "Fewer changes than expected";
+  if(changes.length > all.length) {
+    msg = "More changes than expected";
+  }
+  assert.equal(changes.length, all.length, msg);
+
+  // Because the changes handed to us in expected aren't going to have the same
```
joshuafcole: Good docs :+1:

joshuafcole: NOTE: This algorithm is greedy. If we're convinced there is never a situation where two actual records might satisfy both expected records in one direction and only one in the other that's fine, otherwise this can get us into subtle bug territory. It is simpler than the previous system was though. I think it's fine for now, with a caveat emptor comment.

test/util.ts
```diff
@@ -0,0 +1,119 @@
...
+      if(!found) assert.fail("No match found for: " + actual.toString());
+      else assert.pass("Found match for: " + actual.toString());
+    }
+  }
+}
+
+export function time(start?:any): number | number[] | string {
+  if ( !start ) return process.hrtime();
+  let end = process.hrtime(start);
+  return ((end[0]*1000) + (end[1]/1000000)).toFixed(3);
```
joshuafcole: isn't this `(end[1]/1000000)` off by an order of magnitude? or is lack of sleep getting to me?

> The process.hrtime() method returns the current high-resolution real time in a [seconds, nanoseconds] tuple Array.

1ms = 10e6ns

test/util.ts
ibdknox: 👍

test/util.ts
ibdknox: this is converting seconds into ms (s * 1000) and nanoseconds into ms (ns / 1 000 000). For every million ns we have 1 ms, so it should be ns divided by a million. That's what's happening here as far as I see.

test/util.ts
ibdknox: I don't think this is greedy. It always keeps a list of all the people that could match, so if you have two possible matches that end up being exactly identical, meaning that every value of every field is the same at the same time (which I don't think is possible in our current semantics?), you could potentially end up not having resolved which is which, but at least proven that possibilities exist.

We only reduce the set of possibilities when we've discovered that some actual reduces the possible matches to only one. If we're worried about that happening to multiple somehow, we could keep a lookup of fully resolved guys and make sure the match isn't already taken. I *think* that'd prevent any false positives/negatives that could result from that weirdness.

src/runtime/runtime.ts
ibdknox: Originally I thought the fields needed to be correct and available to the constructor, so I was passing the information we needed down the line. That's obviously not true, so I'll move that logic into DSL-land instead.

test/util.ts
ibdknox: made that change as well, we should be covered now

Issue #734, opened 2017-01-12
CSS libs integration


wilk:
This issue is derived from https://github.com/witheve/Eve/issues/729

I tried to include a whole library like [bulma](http://bulma.io/) but it changed the entire IDE layout and style: https://gist.github.com/anonymous/421dd499f02a4b3d2d35e0589f4ec06c
Is there a way to add a CSS libs without breaking the IDE?

A proposal can be wrapping the IDE preview section into an iframe, isolating the CSS of the application from the IDE.

Others suggestions?

cmontella:
Yeah, I think an iframe is the way to go here. Did you want to take up this project yourself?

wilk:
@cmontella 

Yeah, I'm working on it ;)
Thanks!

cmontella:
Great, keep us posted!

PR #762, opened 2017-01-30
Feature/preview app iframe

wilk:
wrapped preview app box into a custom iframe, related to https://github.com/witheve/Eve/issues/734


cmontella:
Hi Wilk, thanks for taking the initiative on this. I found that the addition of the iframe breaks embedded CSS blocks. For example, this program does not display correctly in the iframe branch: http://play.witheve.com/#gist:06c780db535d63cf104f75f42993aa16-root-foo.eve

(don't mind the first empty CSS block, it's a known bug that the contents of the first CSS block don't get injected into the page )

wilk:
@cmontella yep, you're right: now CSS blocks are supported correctly -> https://github.com/witheve/Eve/pull/762/commits/a092db097fc588656f375ec0c1752fb412f5a00a

PR #770, opened 2017-02-02
BinaryFlow and AntiJoin nodes, hookup DSL not

ibdknox:
This change adds an AntiJoin node type that serves as the basis for blocks that contain a logical negation in them. It also hooks up the DSLNot node that we created and adds a simple test to show what using not looks like.

### DIFF
--- a/src/runtime/dsl.ts
+++ b/src/runtime/dsl.ts
@@ -8,7 +8,7 @@ declare var Proxy:new (obj:any, proxy:any) => any;
 declare var Symbol:any;
 
 import {RawValue, Register, isRegister, GlobalInterner, Scan, IGNORE_REG, ID,
-        InsertNode, Node, Constraint, FunctionConstraint} from "./runtime";
+        InsertNode, Node, Constraint, FunctionConstraint, Change} from "./runtime";
 import * as runtime from "./runtime";
 import * as indexes from "./indexes";
 
@@ -298,6 +298,8 @@ class DSLRecord {
 // DSLBlock
 //--------------------------------------------------------------------
 
+type DSLCompilable = DSLRecord | DSLFunction;
+
 class DSLBlock {
   records:DSLRecord[] = [];
   variables:DSLVariable[] = [];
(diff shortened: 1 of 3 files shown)


src/runtime/dsl.ts
```diff
@@ -531,8 +530,28 @@ class DSLBlock {
...
     // @TODO: Once we start having aggregates, we'll need to do some stratification here
     // instead of just throwing everything into a single JoinNode.
-    nodes.unshift(new runtime.JoinNode(constraints))
+    let join:Node = new runtime.JoinNode(constraints);
+
+    // @NOTE: We need to unify all of our sub-blocks along with ourselves
+    //        before the root node can allocate registers.
+    for(let not of this.nots) {
+      not.compile(items);
+      let notJoinNode = not.block.nodes[0];
```
joshuafcole: We should note that this isn't sufficient in the post-aggregate world too.

src/runtime/runtime.ts
```diff
@@ -1243,6 +1243,103 @@ export class InsertNode implements Node {
...
+  abstract onLeft(index:Index, prefix:ID[], transaction:number, round:number, results:Iterator<ID[]>):void;
+  abstract onRight(index:Index, prefix:ID[], transaction:number, round:number, results:Iterator<ID[]>):void;
+}
+
+export class AntiJoin extends BinaryFlow {
+  leftIndex = new IntermediateIndex();
+  rightIndex = new IntermediateIndex();
+  keyFunc:KeyFunction;
+
+  constructor(public left:Node, public right:Node, public keyRegisters:Register[]) {
```
joshuafcole: What happens when not contains an agg here?

src/runtime/dsl.ts
```diff
@@ -531,8 +530,28 @@ class DSLBlock {
     }
     // @TODO: Once we start having aggregates, we'll need to do some stratification here
     // instead of just throwing everything into a single JoinNode.
-    nodes.unshift(new runtime.JoinNode(constraints))
+    let join:Node = new runtime.JoinNode(constraints);
+
+    // @NOTE: We need to unify all of our sub-blocks along with ourselves
+    //        before the root node can allocate registers.
+    for(let not of this.nots) {
+      not.compile(items);
```
joshuafcole: We should address the way nots embed their parents logic here in a comment

ibdknox: 👍

src/runtime/dsl.ts
ibdknox: 👍

src/runtime/runtime.ts
ibdknox: I'm going to add a LinearFlow node that allows us to stack these guys up arbitrarily.

PR #771, opened 2017-02-03
Choose and union!

ibdknox:
This change adds both choose and union nodes, as well as hooking choose and union up in the dsl. Still need to add comments about how the flows I added here work as there are some subtleties that need to be documented.

### DIFF
--- a/src/runtime/dsl.ts
+++ b/src/runtime/dsl.ts
@@ -8,7 +8,7 @@ declare var Proxy:new (obj:any, proxy:any) => any;
 declare var Symbol:any;
 
 import {RawValue, Register, isRegister, GlobalInterner, Scan, IGNORE_REG, ID,
-        InsertNode, Node, Constraint, FunctionConstraint, Change} from "./runtime";
+        InsertNode, Node, Constraint, FunctionConstraint, Change, concatArray} from "./runtime";
 import * as runtime from "./runtime";
 import * as indexes from "./indexes";
 
@@ -62,7 +62,7 @@ function isRecord(a:any): a is DSLRecord {
 // DSLVariable
 //--------------------------------------------------------------------
 
-type DSLVariableParent = DSLFunction|DSLRecord;
+type DSLVariableParent = DSLFunction|DSLRecord|DSLLookup;
 type DSLNode = DSLFunction|DSLRecord|DSLVariable|RawValue;
 
 type DSLValue = RawValue|Register;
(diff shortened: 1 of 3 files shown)


src/runtime/dsl.ts
```diff
@@ -400,6 +413,24 @@ class DSLBlock {
     this.nots.push(not);
```
joshuafcole: All these `this`'s should  be `active`s

src/runtime/runtime.ts
```diff
@@ -1300,6 +1301,51 @@ abstract class BinaryFlow implements Node {
   abstract onRight(index:Index, prefix:ID[], transaction:number, round:number, results:Iterator<ID[]>):void;
 }
 
+export class BinaryJoinRight extends BinaryFlow {
```
joshuafcole: Let's stick a comment in on this guy

src/runtime/runtime.ts
```diff
@@ -1323,19 +1369,86 @@ export class AntiJoin extends BinaryFlow {
...
-      let result = copyArray(prefix, "AntiJoinResult");
+      let result = copyArray(leftPrefix, "AntiJoinResult");
       result[result.length - 2] = upperBound;
       result[result.length - 1] = count * leftCount * -1;
       results.push(result);
     }
+  }
+}
 
+export class AntiJoinPresovledRight extends AntiJoin {
```
joshuafcole: Presovled

joshuafcole: Let's also note why we've got this guy.

src/runtime/dsl.ts
```diff
@@ -400,6 +442,31 @@ class DSLBlock {
...
+    return node;
+  }
+
+  union = (...branches:(() => any)[]) => {
+    let node = new DSLUnion(branches, this.program);
+    this.unions.push(node);
+    for(let result of node.results) {
+      this.registerVariable(result);
+    }
+    return node.results[0];
```
joshuafcole: Lets add a note here about this not actually supporting multi-field returns.

src/runtime/runtime.ts
```diff
@@ -1310,32 +1394,108 @@ export class AntiJoin extends BinaryFlow {
...
+      let result = copyArray(leftPrefix, "AntiJoinResult");
       result[result.length - 2] = upperBound;
       result[result.length - 1] = count * leftCount * -1;
       results.push(result);
+      debug("                    ->", key, count, leftCount, result[result.length - 1])
+    }
+  }
+}
+
+export class AntiJoinPresovledRight extends AntiJoin {
```
joshuafcole: This guy is still Presovled

PR #772, opened 2017-02-07
Runtime levels

joshuafcole:


src/runtime/dsl.ts
```diff
@@ -527,6 +579,20 @@ class DSLBlock {
     }
   }
 
+  precompile() {
+    this.program.contextStack.push(this);
+
+    for(let not of this.nots) {
```
joshuafcole: Copypasta bug

src/runtime/dsl.ts
```diff
@@ -575,90 +641,243 @@ class DSLBlock {
...
+      let constraints = [];
+      for(let toCompile of items) {
+        if(!toCompile) continue;
+        let compiled = toCompile.compile();
+        if(!compiled) continue;
+        for(let item of compiled) {
+          if(item instanceof Scan || item instanceof FunctionConstraint) {
+            constraints.push(item);
+          } else {
+            nodes.push(item as Node);
```
joshuafcole: We should be  ignoring outputs here

src/runtime/dsl.ts
```diff
@@ -575,90 +641,243 @@ class DSLBlock {
...
         }
       }
     }
-    // @TODO: Once we start having aggregates, we'll need to do some stratification here
-    // instead of just throwing everything into a single JoinNode.
-    let join:Node = new runtime.JoinNode(constraints);
+    // go through all the functions, chooses, and unions to see if they rely on
+    // a register that has been leveled, if so, they need to move to a level after
+    // the provider's heighest
+    concatArray(items, this.cleanFunctions);
```
joshuafcole: We need to level nots as well here.

PR #774, opened 2017-02-08
Runtime aggregates

ibdknox:
This PR adds basic commutative aggregates to the runtime, using `sum` as the motivating example. Right now, aggregates are going to be pretty slow and do a lot of scary crossing at the end that we will ultimately need to find a way around, but this will work for the moment.

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -69,6 +69,16 @@ export function copyArray(arr:any[], place = "unknown") {
   return arr.slice();
 }
 
+export function copyHash(hash:any, place = "unknown") {
+  if(!ALLOCATION_COUNT[place]) ALLOCATION_COUNT[place] = 0;
+  ALLOCATION_COUNT[place]++;
+  let neue:any = {};
+  for(let key of Object.keys(hash)) {
+    neue[key] = hash[key];
+  }
+  return neue;
+}
+
 export function concatArray(arr:any[], arr2:any[]) {
   let ix = arr.length;
   for(let elem of arr2) {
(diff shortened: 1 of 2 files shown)


src/runtime/runtime.ts
```diff
@@ -1578,6 +1578,223 @@ export class ChooseFlow implements Node {
...
+    let prefixRound = prefix[prefix.length - 2];
+    let prefixCount = prefix[prefix.length - 1];
+    let delta = 0;
+    let found = this.groups[group];
+    if(!found) {
+      found = this.groups[group] = {result: []};
+    }
+    let counts = found[projection] || [];
+    let totalCount = 0;
+    for(let count of counts) {
```
ibdknox: this should be the count up to this round.

src/runtime/runtime.ts
```diff
@@ -1578,6 +1578,223 @@ export class ChooseFlow implements Node {
...
+    if(delta < 0) op = this.remove;
+
+    let groupStates = this.groups[group].result;
+    let currentState = groupStates[prefixRound];
+    if(!currentState) {
+      // otherwise we have to find the most recent result that we've seen
+      for(let ix = 0, len = Math.min(groupStates.length, prefixRound); ix < len; ix++) {
+        let current = groupStates[ix];
+        if(current === undefined) continue;
+        currentState = current;
```
ibdknox: this needs to be a copy

src/runtime/runtime.ts
```diff
@@ -1578,6 +1578,223 @@ export class ChooseFlow implements Node {
...
+
+  abstract add(state:any, resolved:RawValue[]):any;
+  abstract remove(state:any, resolved:RawValue[]):any;
+  abstract getResult(state:any):RawValue;
+  abstract newResultState():any;
+
+}
+
+export class SumAggregate extends AggregateNode {
+  add(state:any, resolved:RawValue[]):any {
```
ibdknox: state can be something more specific for the implementation of an agg

PR #775, opened 2017-02-08
Runtime watchers

joshuafcole:
A little bit of cleanup on the watcher interface


src/watchers/watcher.ts
```diff
@@ -30,3 +29,19 @@ export class Watcher {
 
   setup() {}
 }
+
+
+// If we're running on the machine, we can autoload all the watchers
+// for you.  For the browser, we'll still need to build an explicit
+// bundle of watchers.
+// @TODO: We should also import modules with the `eve-watcher`
+//        attribute in their package.json.
```
ibdknox: As is, we won't load the HTML watcher in the browser.

PR #777, opened 2017-02-16
commit and removal

ibdknox:
This adds commit and removal nodes to the runtime and the DSL.


joshuafcole:
In the fix-test-fix cycle for puzzling out the incremental commit logic this has undergone a trial by fire and proved to be the wrong solution. The new logic is in the runtime-commit branch.

PR #785, opened 2017-03-01
Runtime performance

ibdknox:
This includes a bunch of performance fixes to get flappy sailing at 60fps.


src/watchers/dom.ts
```diff
@@ -60,7 +60,7 @@ export abstract class DOMWatcher<Instance extends ElemInstance> extends Watcher
     for(let curIx = 0; curIx < parent.childNodes.length; curIx++) {
       let cur = parent.childNodes[curIx] as Instance;
       if(cur === child) continue;
-      if(cur.__sort !== undefined && cur.__sort > at) {
+      if(cur.__sort !== undefined && at !== undefined && cur.__sort > at) {
```
joshuafcole: Note to self: this logic doesn't group all the undefined's together.

PR #787, opened 2017-03-03
mouseenter and mouseleave events

shamrin:
Usage:

1. add `html/onmouseenter` or `html/onmouseleave` tag to an element
2. listen to `html/event/mousenter` or `html/event/mouseleave`

`hover.ts` example:

untitled

~~**Open question.** Do we need to remove the `mouseenter`/`mouseleave` records, like [we do with clicks][1]?~~

[1]: https://github.com/witheve/Eve/blob/5538d8ebd842ccb0030991e03338afacae659531/src/watchers/dom.ts#L134-L142


shamrin:
I've updated the PR for the recent `refactor-runtime` branch changes.

src/programs/hover.ts
```diff
@@ -0,0 +1,65 @@
...
+    let elem = find("html/div");
+    return [elem.add("tag", "html/element").add("tagname", "div")];
+  })
+  .block("Translate elements into html", ({find, record}) => {
+    let elem = find("html/style");
+    return [elem.add("tag", "html/element").add("tagname", "style")];
+  });
+
+prog.test(1, [
+  [1, "tag", "clock-timer"],
```
joshuafcole: Since you're not using a timer, you can remove these first 3 EAVs

src/programs/hover.ts
```diff
@@ -0,0 +1,65 @@
...
+  .block("Translate elements into html", ({find, record}) => {
+    let elem = find("html/div");
+    return [elem.add("tag", "html/element").add("tagname", "div")];
+  })
+  .block("Translate elements into html", ({find, record}) => {
+    let elem = find("html/style");
+    return [elem.add("tag", "html/element").add("tagname", "style")];
+  });
+
+prog.test(1, [
```
joshuafcole: We can use `prog.inputEavs([
  [1, "tag", "dummy"]
])`

to kick things off here. Functions basically the same, but doesn't require you to choose an appropriate transaction number and doesn't spew debug info to the console.

src/programs/hover.ts
```diff
@@ -0,0 +1,65 @@
...
+  [1, "tag", "clock-timer"],
+  [1, "tag", "system/timer"],
+  [1, "resolution", 1000],
+  [3, "tag", "main"],
```
joshuafcole: Having at least one EAV input is unfortunately necessary right now as it seems you've discovered. We nshould figure out a proper way to start a program without having to insert a bogus input. @ibdknox

src/watchers/html.ts
```diff
@@ -59,6 +60,71 @@ class HTMLWatcher extends DOMWatcher<Instance> {
 
       this._sendEvent(eavs);
     });
+
+    let {program:me} = this;
```
joshuafcole: For consistency, let's chain these calls instead of putting them into the `me variable.

E.g.

```
this.program
  .watch(...)
  .asObjects(...)
  .watch(...)
  // ...
```

joshuafcole:
As a side note: We've discussed whether or not some, all, or no events should be opt in (as you've done with these two). We're somewhat divided on the decision but we've decided that's easier for users if we're explicit as you've written it from the start and then make it implicit later if we need to.

shamrin:
@joshuafcole Thank you for the review! I'll fix the stuff you've raised.

Having to opt-in for `mouseenter`/`mouseleave` events was simply the easiest way to implement. I can think of two alternatives:

1. Track all mouse movements with `window.addEventListener('mousemove', …)`. Calculate all elements' sizes and positions (`getBoundingClientRect`?) and do hit test calculations for all of them. Raise the appropriate events.
2. Find the blocks that search for `html/event/mouseover` and attach the `addEventListener` for the mentioned elements.

The first approach is probably too slow and complicated. Second approach is better, but I don't know how to analyze blocks in this way.

shamrin:
@joshuafcole   I've fixed the style issues.

In addition, I've tried to make a more complex example, but I coudn't make it work. I wanted to postpone lower square disappearance for 1 second. And also allow moving the mouse focus to the lower square. I've pushed [`hover2.ts`][1] to [hoverbugs branch][2].

[1]: https://github.com/shamrin/Eve/blob/hoverbugs/src/programs/hover2.ts
[2]: https://github.com/shamrin/Eve/tree/hoverbugs

untitled

Problems:

1. expiration.time should go to `Infinity` on each `mouseenter`, but it works only for the first time
2. lower square never goes away. It seems [one of my bind blocks][3] doesn't work correctly.

[3]: https://github.com/shamrin/Eve/blob/hoverbugs/src/programs/hover2.ts#L69-L78

joshuafcole:
Thanks for the cleanup, looks great!

As to `hover2`, The runtime is actually working there, but the new JS syntax is a little clunky. Since you only ever want there to be a single `time` on the expiration, you want to do something like `expiration.time := Infinity` In the JS DSL, `expiration.add("time", Infinity)` is actually `expiration.time += Infinity`, which means you're adding multiple times to the expiration object. To get `:=` behavior in the new syntax, use `expiration.remove("time").add("time", Infinity)`. If you do, then it works perfectly. :)

hover2

As to your addition of `timestamp` in the `hoverbugs` branch, I'd  be happy to pull that in as a separate PR if you're so inclined.

shamrin:
@joshuafcole Thank you for help! I was suspecting that I had multiple `time`, but I didn't know how to check for that. I've tried:

```
let expiration = find("expiration");
let count = gather(expiration).count();
```

But it's always `1`. On the other hand, `gather(expiration.time).count()` is always `0`.

And I'm still surprised that I had only one `expiration.time: …` line on a page. I was expecting to have multiple lines, for each `time`.

Regarding timestamps, I've submitted PR #790.

joshuafcole:
Yeah, let's deconstruct each issue since I think they provide some insight into awkwardness of the JS syntax.

To start with, the aggregate: `gather(expiration).count()` is counting the number of records tagged `#expiration` as written. Since you only create a single `#expiration` once (at the end in inputEavs) and then add/remove times to it, this will only ever be 1. Instead, you could count the number of times with `gather(expiration.time).count()` or the number of record and time pairs with `gather(expiration, expiration.time).count()`.

Unfortunately, given the way the debug block is written:
```
record("html/div", {sort: -99})
          .add("text", "timer: " + timer.timestamp),
```

Creates a single div element who's only unique identity is that his `sort` is `-99`. No matter how many attributes you attach to this later, there will still only be this one div. In the old renderer, we'd have silently joined each text with a comma and inserted the whole string. I've decided not to do something like that at this level because it requires an ordering that we lack information to provide. If a div has multiple texts, which comes first? What happens when a new one is inserted? If you refresh with the state persisted, shouldn't the order be somehow consistent? Etc. All of those expectations would be violated by the old system.

As to why the new one didn't error -- that's a separate and interesting question. But as it stands it seems to erroneously just have shown the most recently assigned time.

PR #790, opened 2017-03-10
add timestamp to #system/timer

shamrin:


joshuafcole:
@cmontella Make sure we note in the future docs that this is in milliseconds rather than seconds (like the unix timestamp traditionally is).

joshuafcole:
Thanks!

PR #791, opened 2017-03-11
Runtime compiler watcher

ibdknox:
This adds a watcher that can compile joins/outputs from EAVs in the database. It also adds `BlockChangeTransaction`, which handles correctly adding and removing blocks from an existing program by running the join nodes with an empty input.


src/runtime/runtime.ts
```diff
@@ -1649,61 +1650,69 @@ export class JoinNode extends Node {
...
-        if(isIncluded) {
-          let valid = constraint.applyInput(input, prefix);
-          // debug("        included", printConstraint(constraint));
-          // If any member of the input constraints fails, this whole combination is doomed.
-          if(valid === ApplyInputState.fail) {
-            shouldApply = false;
-            break;
+    this.inputCount = input.count;
+
+    if(input === BLOCK_REMOVE) {
```
joshuafcole: We should add a comment for add and remove like "to remove a block we run with no applied inputs and a negative count, which gives us retractions of all the changes the block produces."

src/watchers/compiler.ts
```diff
@@ -0,0 +1,309 @@
...
+    let item = items[blockID];
+    let {name, constraints, type} = item;
+    let compile:CompilationContext = {variables: {}};
+    let flow:LinearFlow;
+    if(type === "block") {
+      flow = new LinearFlow(() => []);
+    } else if(type === "watch") {
+      flow = new WatchFlow(() => []);
+    } else {
+      flow = new LinearFlow(() => []);
```
joshuafcole: I guess this third guy is a commit?

src/watchers/compiler.ts
```diff
@@ -0,0 +1,309 @@
...
+        }
+        found.attributes.push([attribute, value]);
+        this.queue(block);
+      }
+      for(let key in removes) {
+        let {block, id, record, attribute, value} = removes[key];
+        let found = items[id];
+        if(!found) { continue; }
+
+        found.attributes.filter(([a, v]:RawValue[]) => a !== attribute || v !== value);
```
joshuafcole: This is dead code. Is it meant to do something?

src/watchers/compiler.ts
```diff
@@ -0,0 +1,309 @@
...
```
joshuafcole: Also dead

PR #795, opened 2017-03-15
Support all DOM events

shamrin:
* [x] subscribe to events by `[… on: "mouseenter"]`
* [x] change `[#html/event/…]` to `[#dom/event, event: …]`
* [x] add `events.ts` example
* [ ] remove listeners when removing `on` attribute
* [ ] change old `[#html/event/click]` to `[#dom/event, event: "click"]`
* [ ] get access to `<input>` value (likely in a separate PR)
* [ ] `event.remove("tag")` → `event.remove()` once PR #797 is part of `refactor-runtime` branch

PR is not finished because I'm not sure about `[#dom/event, event: "click"]` part and about the explicit subscription. Comments?

untitled

### DIFF
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -55,3 +55,4 @@ function doIt() {
 doIt();
 // import "./programs/flappy";
 // import "./programs/hover";
+// import "./programs/events";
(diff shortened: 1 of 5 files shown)


cmontella:
We have to sort of come up with a convention here. In the eve syntax, we have 

```
search @event
  [#click]
```

Which would make sense to become `[#dom/event/click]`

I think we should resist the temptation to be too highly specific with tags though, keeping them shallow and getting more specific with attributes. "dom/event" is probably specific enough a tag, so I agree  with the change you've made.

shamrin:
Sorry for late reply! I'm closing this, because the issue was fixed by PR #823 and other changes.

PR #797, opened 2017-03-16
add free AV remove and a test

ibdknox:
Add the ability to call `foo.remove()` to remove all the attributes attached to a record, thus removing the record

### DIFF
--- a/src/programs/compiler.ts
+++ b/src/programs/compiler.ts
@@ -59,7 +59,7 @@ prog.inputEavs([
   ["v2Record", "tag", "eve/compiler/output"],
   ["v2Record", "tag", "eve/compiler/remove"],
   ["v2Record", "record", "v1"],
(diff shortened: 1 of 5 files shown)


shamrin:
Does this mean we can now do `.remove()` instead of `.remove("tag")` in `dom.ts` watcher?

joshuafcole:
Yep, that's correct. Now we don't have to orphan garbage facts like we
previously were. It's equivalent to `event := none`

On Thu, Mar 16, 2017 at 3:21 PM Alexey Shamrin <notifications@github.com>
wrote:

Does this mean we can now do .remove() instead of .remove("tag") in dom.ts
watcher?

—
You are receiving this because you modified the open/close state.


Reply to this email directly, view it on GitHub
<https://github.com/witheve/Eve/pull/797#issuecomment-287208860>, or mute
the thread

PR #804, opened 2017-03-25
Larger String Standard Library

christopherdumas:
Here are some useful string library functions for Eve (all present in the JS standard library, making my job easier).

### DIFF
--- a/examples/untitled.eve
+++ b/examples/untitled.eve
@@ -1,8 +1,0 @@
-# Untitled
-```
-search
-  foo = suffix-of[text: "foo bar baz", suffix: "baz"]
-
-bind @view
-  [#value value: foo]
-```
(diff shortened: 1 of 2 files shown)


cmontella:
With the 0.3 runtime coming out, the way functions are defined has changed, but they're much simpler. The stdlib functions are defined here: https://github.com/witheve/Eve/blob/master/src/runtime/stdlib.ts

I'm going to close this PR, and if you want, please re-implement your functions in the new way. Let me know if you'd like me to handle it; I'll just use your PR here as a guideline.

christopherdumas:
O thank God the new way look so much better. My school year (although wrapping up) is still really busy so I probably won't get around to porting this PR until the summer. However I would still like to do it.

PR #805, opened 2017-03-27
Fix aggregates in choose

ibdknox:
simplify results ownership, add AggregateOuterLookup node, get all the aggs in chooses tests passing

### DIFF
--- a/src/runtime/runtime.ts
+++ b/src/runtime/runtime.ts
@@ -219,6 +219,22 @@ export class Iterator<T> {
     if(this.ix < this.length) return this.array[this.ix++];
     return;
   }
+
+  iter():ReadOnlyIterator<T> {
+    return new ReadOnlyIterator(this.array, this.length);
+  }
+}
+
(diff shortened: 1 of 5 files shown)


src/runtime/runtime.ts
```diff
@@ -2610,30 +2599,110 @@ export class MergeAggregateFlow extends BinaryJoinRight {
...
     return true;
   }
+
+  onLeft(context:EvaluationContext, prefix:Prefix, transaction:number, round:number, results:Iterator<Prefix>):void {
+    let key = this.keyFunc(prefix);
+    let exists = this.leftIndex.has(key);
+    this.leftIndex.insert(key, prefix);
+    let afterExists = this.leftIndex.has(key);
+    let diffs = this.rightIndex.iter(key, 0);
+    if(exists && afterExists || !diffs) return;
```
joshuafcole: Is it ever possible to not exist and continue not existing here?

joshuafcole: impossible :+1:

PR #824, opened 2017-04-19
Test that shows how to emulate LIMIT queries

shamrin:
**I'm not sure this PR should be merged.**

I wanted to do LIMIT queries in Eve. First I've tried `gather(…).count() <= 5`. And I was surprised to get zero results in some cases. I've even thought it's a bug in Eve. But it's not, because `0 <= 5` (obviously!).

Then I've studied the `test/` folder. And finally I've come up with: `let pos = gather(…).sort(); pos <= 5`. ~~Which works great.~~ (see my comments to this PR)

I view `test/` folder as the best example on how to use Eve query language. That's why I'm adding "limit query" test there.

(I've also renamed some blocks in `test/aggregate.ts`.)

### DIFF
--- a/test/aggregate.ts
+++ b/test/aggregate.ts
@@ -203,12 +203,14 @@ test("Aggregate: multi-direction sort", (assert) => {
   assert.end();
 });
 
-test("Aggregate: limit query with sort and `pos <=`", (assert) => {
+test("Aggregate: limit query with sort, count and range", (assert) => {
   let prog = new Program("test");
-  prog.bind("limit query with sort and `pos <=`", ({find, gather}) => {
+
+  prog.bind("limit query with sort, count and range", ({find, gather, lib:{math}}) => {
     let person = find("person");
     let pos = gather(person.name, person).sort();
-    pos <= 2;
+    let upper = math.min(gather(pos).count(), 2);
+    pos == math.range(1, upper);
     return [
       person.add("pos", pos)
     ];
--- a/test/aggregate.ts
+++ b/test/aggregate.ts
@@ -55,7 +55,7 @@ test("Aggregate: Count in choose", (assert) => {
 test("Aggregate: direction-less sort", (assert) => {
 
   let prog = new Program("test");
-  prog.bind("test count in choose", ({find, choose, gather, record}) => {
+  prog.bind("test direction-less sort", ({find, choose, gather, record}) => {
     let person = find("person");
     let pos = gather(person.name).sort();
     return [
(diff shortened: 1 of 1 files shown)


shamrin:
Now that I think about it, `let pos = gather(…).sort(); pos <= 5` could in theory produce zero results. What do you think? If yes, then this PR really becomes a questions: how do I do LIMIT queries in Eve?

shamrin:
I've updated the test to *also* use `count()`, `math.range` and `math.min`. Seems somewhat verbose, but now looks certain now. (And I was surprised to see Eve handling this query without problems.)

cmontella:
Right, limit in Eve can be defined with sort and a filter. Here it is working in the v0.2.3 branch: http://play.witheve.com/#gist:905669d493651fd15fe5a465706b8258-examples-untitled.eve

shamrin:
Thank you!

I'm confused that limiting stops working when I replace `pos <= 2` with `gather(person).count() <= 2`.

Works:

```javascript
  prog.bind("limit query with sort", ({find, gather}) => {
    let person = find("person");
    let pos = gather(person.name, person).sort();
    pos <= 2;
    return [
      person.add("pos", pos)
    ];
  });
```

Doesn't work:

```javascript
  prog.bind("limit query with count", ({find, gather}) => {
    let person = find("person");
    let pos = gather(person.name, person).sort();
    gather(person).count() <= 2;
    return [
      person.add("pos", pos)
    ];
  });
```

(https://github.com/shamrin/Eve/blob/count-vs-sort/test/aggregate.ts#L206-L234)

Why does the first work and the second one doesn't? I came to conclusion that Eve was right to choose zero results in the second example, because all conditions hold. What if first example working is simply an accident?

cmontella:
The problem with the second example isn't really doing what you want. You want "give me at most 2 people". But the second example says "Give me only 1 or 2 people".

If there are 3 people, then `gather(person).count()` is equal to 3. But your constraint of `<= 2` invalidates the entire block in this case. As you know, Eve is referentially transparent, so we can just evaluate and substitute the first part of the expression: `3 <= 2` Which makes things much more obvious.

joshuafcole:
Hey @shamrin,

Aggregates create 0 or 1 result if you don't specify a grouping (if you do, you get one result per group). Because of this, `gather(person).count()` can only ever return a single number. If that number is less than or equal to two, all the people will pass, otherwise none of them will. When you want to do a limit query, you need to use sort (which provides one result per each unique projection).

Therefore, `gather(person).sort() <= 2` should work fine. Just keep in mind that Eve is 1-indexed, so that should yield 0, 1, or 2 results depending on the number of people you have in the system.

joshuafcole:
I've also gone ahead and added a test case to demonstrate limits like you requested. Please look it over and let me know if it meets your expectations. https://github.com/witheve/Eve/pull/825

I'd also like to accept your renaming changes here if you can update the PR to remove or correct the new test.

shamrin:
> If that number is less than or equal to two, all the people will pass, otherwise none of them will.

Thank you, now it makes perfect sense! For some reason I couldn't understand that zero results is the same as having block that doesn't match.

I've updated this PR. Only renaming changes are left.

PR #825, opened 2017-04-19
Add limit test

joshuafcole:
@shamrin recommended that we add a test case showing how limits work using aggregates.

@shamrin the GH UI doesn't let me request a review from non-maintainers, but please look this over and let me know if it's in line with what you expected.


shamrin:
Looks good to me! :+1:

PR #827, opened 2017-04-20
Switch off from deprecated node-uuid package

shamrin:
This replaces [deprecated `node-uuid`](https://www.npmjs.com/package/node-uuid) package with [`uuid`](https://www.npmjs.com/package/uuid).

As a bonus, it shaves 92K from my app webpack bundle:

|   Filename      | before           | after  |
| :------------- |-------------:| -----:|
| main.js     | 1.2M | 649K |
| main.min.js      | 554K      |   271K |
| main.min.js.gz | 164K      |    72K |

It's because `node-uuid` was bringing huge amount of crypto packages with it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,15 +31,14 @@
   },
   "dependencies": {
     "@types/node": "^6.0.41",
-    "@types/node-uuid": "0.0.28",
     "@types/tape": "^4.2.28",
     "chevrotain": "^0.19.0",
     "express": "^4.14.0",
     "falafel": "^2.0.0",
     "javascript-natural-sort": "^0.7.1",
-    "node-uuid": "^1.4.7",
     "setimmediate": "^1.0.5",
-    "typescript": "^2.1.4"
+    "typescript": "^2.1.4",
+    "uuid": "^3.0.1"
   },
   "devDependencies": {
     "faucet": "0.0.1",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "@types/node": "^6.0.41",
     "@types/tape": "^4.2.28",
+    "@types/uuid": "^2.0.29",
     "chevrotain": "^0.19.0",
     "express": "^4.14.0",
     "falafel": "^2.0.0",
--- a/src/programs/perfReport.ts
+++ b/src/programs/perfReport.ts
@@ -1,7 +1,7 @@
 //---------------------------------------------------------------------
 // Performance Report
 //---------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


cmontella:
I think we need to include the uuid typedefs still: https://www.npmjs.com/package/@types/uuid

@jjosh is that the case?

shamrin:
@cmontella I've tried `@types/uuid`, but it doesn't work for `import * as uuid "uuid/v4". I didn't want to spend time figuring out how to fix it. That's why I've added declaration to `src/types.d.ts`.

joshuafcole:
This looks good. I hadn't realized how heavy-weight `node-uuid` was. There was a minor issue with the starter, but I've handled it in a companion PR: https://github.com/witheve/eve-starter/pull/8.

As @cmontella mentioned though, let's go ahead and use the official typings file. The reasojn you're getting that error is because CommonJS and ES6 modules aren't 100% compatible, Typescript has provided a way to circumvent this problem (at least for now) by importing the module as `import uuid = require("uuid/v4");` instead of `import * as uuid from "uuid/v4"`.

Once that's done, we'll include this in the release of `preview3` next week. Thanks for the help!

shamrin:
No, the problem is not related to CommonJS vs ES6 incompatiblity. If I do `npm install --save @types/uuid` and then import this way:

```javascript
import uuid = require("uuid/v4");
```

TypeScript still gives me:

```
src/watchers/watcher.ts(6,23): error TS7016: Could not find a declaration file for module 'uuid/v4'. '/Users/user/src/Eve/node_modules/uuid/v4.js' implicitly has an 'any' type.
```

The following works though:

```javascript
let uuid = require("uuid/v4")`
```

But only because `require` function return type is `any`. And TypeScript stops checking what is passed to `require`.

I think the problem exist because `@types/uuid` package lacks types for `uuid/v4`. It needs to be extended, but it's immediately obvious to me how to do it.

joshuafcole:
Hey @shamrin,

Unfortunately ES6 modules do not allow for directly exporting a function as the export of a module. Unlike CJS, it must be an attribute. Old typing declarations handled this using ambient namespace declarations, but as you say this file doesn't do that either. Luckily the module itself (and the typings) export `v4` as an attribute on the uuid module itself, so let's import as `import {v4 as uuid} from "uuid";`

shamrin:
Thank you, it works! I've updated the PR.

joshuafcole:
I'll hold off on the companion PR for eve-starter until we release `preview3` next week, so that the master branch there will continue working with the current weekly release.

PR #836, opened 2017-06-02
Dont rely on event.buttons for mouse events. Fix #835

### LINKED ISSUE #835: Issues with the click event
The following code demonstrates a few issues I experienced with the click event.

First, the click event fires for all mouse buttons, which was surprising, I would expect the click event to fire just for the left mouse button.

Second, the button attribute is not set on the click event so I can't check for click events where the left button is pressed. 

I found a work around for this. You can use the mouse-down event instead of the click event and the button attribute is set. So in the below code if you change the click event to mouse-down event and add the line clickEvent.button=="left", then the code will work as expected.

```javascript
import {Program} from "witheve";

let prog = new Program("Test Program");
prog.attach("ui");

prog.commit("test",({record,find})=>{
    return [
        record("ui/row","click_target",{text:"test1"}),
        record("ui/row","click_target",{text:"test2"}),
        record("ui/row",{children:[
            record("ui/row",{"text":"output:"}),
            record("ui/row","output")
        ]})
    ]
});

prog.commit("output text click",({record,find})=>{
    let clickTarget=find("click_target");
    let clickEvent=find("html/event/click",{element:clickTarget});      
    return[
        record("ui/row","output").remove("text").add("text",clickTarget.text)
    ]
});

prog.inputEAVs([
  [1, "tag", "turtle"]
]);

joshuafcole:


ibdknox:
We may want to revisit the choice to nuke the mousedown, but this seems reasonable for now.

PR #847, opened 2017-06-16
Better error message for non-fixpoint

cmontella:


src/runtime/runtime.ts
```diff
@@ -3455,14 +3455,15 @@ export class Transaction {
...
-      let change = changes[changeIx];
+    while (changeIx < changes.length) {
+      var change = changes[changeIx];
       tracer.input(change);
       total++;
-      if(total > 10000) {
-        console.error("bad");
-        break;
+      if (total > iterationLimit) {
+          console.error(`Failed to fixpoint after ${iterationLimit} iterations`);
```
joshuafcole: Let's word this such that it's more obvious the program is at fault. Maybe `Error: Program failed to fixpoint after ${iterationLimit} iterations. This is likely due to an unbounded cycle in the program.` or something. Other than that, looks fine.

joshuafcole: let's not add those spaces after keywords like `if` and `while` though, we aren't doing that elsewhere.

PR #854, opened 2017-06-22
Let users watch for mouse-move events

ThatOtherPerson:


joshuafcole:
Hey @ThatOtherPerson, we'd like to work with you to get this in, but since Eve 0.3 isn't really fast enough to receive a continuous slew of mouse move events, we'll need to do a more limited version. To start with, we'd like to bar this feature behind an `#html/listener/mouse-position` tag (just like hover is [1] [2]). The handler will bubble from the target and will only dispatch to Eve if at least one of the elements in the hierarchy has that listener. Additionally, it will only attach event information about the element(s) with that listener to keep volume down.

Once we've done that we'll do some qualitative testing to see if we also need to throttle the events.

Are you interested in spearheading this?

[1]: http://docs.witheve.com/v0.3/handbook/libraries/stdlib/#html 
[2]: https://github.com/witheve/Eve/blob/master/src/watchers/html.ts#L235

ThatOtherPerson:
Yes, sounds good!

ThatOtherPerson:
@joshuafcole I've tried implementing your suggestions, let me know if there's anything more I should fix.

src/watchers/html.ts
```diff
@@ -251,6 +251,51 @@ export class HTMLWatcher extends DOMWatcher<Instance> {
...
+            [eventId, "tag", `html/event/${tagname}`],
+            [eventId, "page-x", event.pageX],
+            [eventId, "page-y", event.pageY],
+            [eventId, "window-x", event.clientX],
+            [eventId, "window-y", event.clientY],
+
+            [eventId, "target", elemId]
+          );
+
+          let current:Element|null = currentTarget;
```
joshuafcole: We should instead do

```
currentTarget = currentTarget.parentNode
```

This way we only have a single loop, and visit each element in the hierarchy once. This will also give us a separate event per element, which gives us the flexibility to add `local-x` and `local-y` attributes to the mose-position event.

src/watchers/html.ts
```diff
@@ -251,6 +251,51 @@ export class HTMLWatcher extends DOMWatcher<Instance> {
...
+          let elemIds = [];
+          while (current && this.isInstance(current)) {
+            if (current.listeners && current.listeners["mouse-position"]) {
+              eavs.push([eventId, "element", current.__element!]);
+            }
+
+            current = current.parentElement;
+          }
```
joshuafcole: Doing the above lets us delete this entire loop.

ThatOtherPerson:
@joshuafcole Thanks for the feedback! The PR has been updated.

src/watchers/html.ts
```diff
@@ -251,6 +251,44 @@ export class HTMLWatcher extends DOMWatcher<Instance> {
...
+      if (!this.isInstance(target)) return;
+
+      let currentTarget:Element|null = target;
+
+      let eavs:(RawEAV|RawEAVC)[] = [];
+
+      while (currentTarget && this.isInstance(currentTarget)) {
+        let elemId = currentTarget.__element!;
+
+        if (currentTarget.listeners && currentTarget.listeners["mouse-position"]) {
```
joshuafcole: Let's use the same name for both the listener and the event. Either `mouse-position` or `mouse-move` seems fine.

joshuafcole:
Sorry for the slow response, between writing an eve-mode for emacs, putting out some fires in 0.2, and preparing for the holidays my time has been fairly well accounted for. I should be much more available after the 4th!

ThatOtherPerson:
No problem!

src/watchers/html.ts
joshuafcole: @ThatOtherPerson This one change and the PR looks good to merge! I've tested it locally and everything is :+1:.

joshuafcole:
Hey @ThatOtherPerson,

Since we're approaching end of life for 0.3, I'm going to close this PR. If you happen to free up after 0.4 is released, ping me and I'd be happy to help you port your work!

PR #855, opened 2017-06-23
rename frames to ticks in the timer

cmontella:


joshuafcole:
Let's bump the version to preview 5 in the package.json as well @cmontella

PR #858, opened 2017-06-26
Add file and console libraries

cmontella:


src/watchers/file.ts
```diff
@@ -0,0 +1,63 @@
...
+                let {readfile, path} = adds[id];
+                fs.readFile(path, {encoding: 'utf-8'}, function(err,data){
+                    if (!err) {
+                        let changes:any[] = [];
+                        changes.push(
+                            [readfile, "contents", data, 1],
+                        );
+                        me.inputEAVs(changes);
+                    } else {
+                        console.log(err);
```
joshuafcole: This should probably drop in an error message of some sort. We haven't fleshed out what the story for #error is yet, but it'd be perfectly fine to have `readfile.error += [code: <err code>, name:, text]` for the program to look for and handle.

src/watchers/file.ts
```diff
@@ -0,0 +1,63 @@
...
+                    }
+                });
+            })
+        })
+
+        // Console log watcher
+        me.watch("print to console log", ({find, record}) => {
```
joshuafcole: As you mentioned, let's split this guy out.

src/watchers/file.ts
```diff
@@ -0,0 +1,63 @@
...
+                            [readfile, "contents", data, 1],
+                        );
+                        me.inputEAVs(changes);
+                    } else {
+                        console.log(err);
+                    }
+                });
+            })
+        })
+
```
joshuafcole: let's add a `#file/write` too? Then I can use this with the websocket watcher for the server backing the editor.

src/watchers/console.ts
```diff
@@ -0,0 +1,53 @@
+import * as fs from "fs";
+import {Watcher} from "./watcher";
+import {ID} from "../runtime/runtime";
+
+class consoleWatcher extends Watcher {
```
joshuafcole: UpperCamel Case `ConsoleWatcher`

src/watchers/console.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    setup() {
+        if(console) {
+            let {program:me} = this;
+
+            me.watch("print to console log", ({find, record}) => {
```
joshuafcole: Let's format these as sentences for consistency. E.g. "Print to console."

src/watchers/console.ts
```diff
@@ -0,0 +1,53 @@
...
+                let log = find("console/log");
+                return [
+                    record({log, text: log.text})
+                ]
+            })
+            me.asObjects<{log:ID, text:string}>(({adds, removes}) => {
```
joshuafcole: asDiffs is a fair bit faster than asObjects. For cases where you only need an id and a single attribute, I recommend using asDiffs instead. E.g.

```
me
  .watch("print to console log", ({find, record}) => {
    let log = find("console/log");
    return [log.add("text", log.text)];
  })
  .asDiffs(({adds}) => {
    for(let [log, _, text] of adds) {
      console.log(text);
    }
  })
```

src/watchers/console.ts
```diff
@@ -0,0 +1,53 @@
...
+                Object.keys(adds).forEach((id) => {
+                    let {log, text} = adds[id];
+                    console.warn(text);
+                })
+            })
+        }
+    }
+}
+
+Watcher.register("console", consoleWatcher);
```
joshuafcole: Missing newline at end of file.

src/watchers/file.ts
```diff
@@ -0,0 +1,84 @@
...
+bind
+  file.error += error
+~~~     
+        `)
+        if(fs.readFile) {
+            me.watch("read a file", ({find, record}) => {
+                let file = find("file/read");
+                return [
+                    record({file, path: file.path, encoding: file.encoding})
```
joshuafcole: I was going to suggest this. :) Let's give it `utf-8` as a default though, e.g.
```
let encoding = choose(() => file.encoding, () => "utf-8");
```

src/watchers/file.ts
```diff
@@ -0,0 +1,84 @@
...
+                ]
+            })
+
+            me.asObjects<{file:ID, path:string, encoding:string}>(({adds, removes}) => {
+                Object.keys(adds).forEach((id) => {
+                    let {file, path, encoding} = adds[id];
+                    fs.readFile(path, {encoding: encoding}, function(err, contents){
```
joshuafcole: `{encoding: encoding}` => `{encoding}`

src/watchers/file.ts
```diff
@@ -0,0 +1,84 @@
...
+                        if (!err) {
+                            let changes:RawEAVC[] = [];
+                            changes.push(
+                                [file, "contents", contents, 1],
+                            );
+                            me.inputEAVs(changes);
```
joshuafcole: This whole branch can be reduced to `me.inputEAVs([[file, "contents", contents]])`. The count is optional and defaults to 1.

src/watchers/file.ts
```diff
@@ -0,0 +1,84 @@
...
+                        } else {
+                            let id = `${file}|error`
+                            let changes:RawEAVC[] = [];
```
joshuafcole: If you type this as `RawEAV` you can omit the counts.

src/watchers/file.ts
```diff
@@ -0,0 +1,84 @@
...
+                return [
+                    record({file, path: file.path, encoding: file.encoding, contents: file.contents})
+                ]
+            })
+            me.asObjects<{file:ID, path:string, contents: string, encoding:string}>(({adds, removes}) => {
+                Object.keys(adds).forEach((id) => {
+                    let {file, path, contents, encoding} = adds[id];
+                    fs.writeFile(path, contents, {encoding: encoding}, function(err){
+                        if (!err) {
+                            console.log(`Write file success: ${contents}`)
```
joshuafcole: Instead of printing, let's set `status: "complete"` on the file record or something similar.

src/watchers/console.ts
```diff
@@ -0,0 +1,47 @@
...
+            this.program
+            .watch("Print to console log.", ({find, record}) => {
+                let log = find("console/log");
+                return [
+                    record({log, text: log.text})
+                ]
+            })
+            .asDiffs(({adds}) => {
+                for(let [log, _, text] of adds) {
+                console.log(text);
```
joshuafcole: Seems like these for loops have all lost their indentation.

PR #859, opened 2017-06-27
Feature/notify watcher

joshuafcole:
Adds the notify watcher, a low effort way to present notices to your users.

notify-demo.eve


cmontella:
Can we put some default styles on them?

PR #864, opened 2017-06-29
make index-of one indexed

cmontella:


joshuafcole:
I'm the one who put the type coercion in there, but I haven't decided yet if that was the right move. It seems equally (or more?) valid to just bail if someone gives us a number as the `text` attribute, but since we don't have any way to convert from string to number or vice versa yet I threw that in. That said, adding `eve/convert[value: "7" to: "number"]` is not hard if we decide it's the right thing to do...

I don't know that we need to slay this dragon now, but it'll become more pressing in 0.4 or if we start accumulating a wide userbase.

joshuafcole:
@cmontella You'll need to look at the editor and ui watchers as well, which both have a single use of `index-of`

cmontella:
I think they are both fine. In each case, you're not using the return value for anything.

PR #867, opened 2017-06-30
Fixed hover text for expand/collapse btn.

maacl:


joshuafcole:
This looks good @maacl, thank you for the contribution. Since we just released on Thursday, it may be a little while until we have accumulated sufficient changes to bump the version to `0.2.6`, but once we do this will be there. :+1:

PR #872, opened 2017-07-10
Add math/ln[] function

cmontella:


joshuafcole:
Javascript's Math.log is actually mathematical ln (log_e). JS also has `log10` in everyone including Edge, but it's not in IE. I'd consider naming this `ln` anyway, no reason to copy JS's weird decisions. Otherwise we could have a generic `log[base, value]`

cmontella:
I think we should have a generic log function eventually, but a ln specific one is good too since it's so common.

PR #883, opened 2017-11-25
Add a function that lets you join strings together

ChrisBrownie55:
The function `string/join` would be very useful to be able to join multiple strings together via a delimiter like a space or comma.


joshuafcole:
Thanks again for your time, but we're going to hold back on trying to maintain 0.3 as well. We'd like to send you a couple of Eve stickers to show our appreciation anyway! Please email me if you're interested.

PR #884, opened 2017-11-25
Update autocomplete sorting

ChrisBrownie55:
Autocomplete now has the ability to **sort matches by the alphabet** as well as the default by length. The way this is done is by adding another value to your autocomplete called `sortby`. If `sortby: "alpha"` then it sorts the matches by the alphabet. Not sure if this works 100%; modification is probably needed.

### DIFF
--- a/src/watchers/ui.ts
+++ b/src/watchers/ui.ts
@@ -403,10 +403,13 @@ export class UIWatcher extends Watcher {
         return [autocomplete.add("match", completion)];
       })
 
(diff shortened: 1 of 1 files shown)


joshuafcole:
In 0.4, we've resolved this issue by allowing the user to specify the sort on options the same way they do for other UI elements. Whatever value is on the 'sort' attribute of the options will be used to determine their order at render time.

PR #885, opened 2017-11-25
Add string/split

ChrisBrownie55:
I'm not sure if that is the right way to do this but maybe you could pass the results back in the form of a record. Not sure how to do that. Or maybe the user could choose the index they want to retrieve from the resulting array from splitting. Overall I think splitting strings would be helpful for when you've joined strings and need to unjoin them.

### DIFF
--- a/src/runtime/stdlib.ts
+++ b/src/runtime/stdlib.ts
@@ -261,6 +261,16 @@ makeMultiFunction({
 //--------------------------------------------------------------------
 
 makeFunction({
+  name: "string/split",
+  args: {text: "string", delimiter: "string"},
+  returns: {result: "string[]"},
+  apply: function(text:string, delimiter:string) {
+    let result = text.split(delimiter);
+    return [result];
+  }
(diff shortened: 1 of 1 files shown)


Joshua-Rowe:
Split was a feature of version 0.2.

Can be found in 0.2 documentation.
http://docs-next.witheve.com/v0.2/handbook/strings/split/

Although, it somehow did not make it into 0.3.

It also may not be worth submitting PRs for this branch right now as 0.4 is on its way and may be a complete, incompatible,  rewrite of the runtime in Rust instead of TS. Not completely sure as it is hard to know since all work in the past months has been secret due to patent concerns.

joshuafcole:
Hey @ChrisBrownie55,

Thanks for doing this! Unfortunately as @Joshua-Rowe pointed out we've decided to skip 0.3 completely. We experimented with a new runtime model which proved unreasonably slow. We're not looking to take additional improvements to it at this time.

